PR #12, opened 2015-01-21
Implementing SecretsWatcher and associated code (Consul Service)

wkimeria:


borderpatrol-core/src/test/scala/com/lookout/borderpatrol/session/SecretsWatcherApiSpec.scala
```diff
@@ -0,0 +1,34 @@
...
+ * Created by wkimeria on 1/15/15.
+ */
+class SecretsWatcherApiSpec extends FlatSpec with Matchers{
+
+  behavior of "SecretWatcherApi"
+
+  def MockConsulService(response: HttpResponse) = new Service[HttpRequest, HttpResponse] {
+    def apply(request: HttpRequest) = {
+      val content =
+        """{'current':{'expiry':{'nanos':100000000000},
```
trane: i think this is out-of-date now (I think expiry now uses "ms" instead of "nano"). wouldn't it make more sense to make content be the encoded Secrets instead of hard-coding the string -- it would have handled this conversion for you and the bug wouldn't have come up.

borderpatrol-core/src/main/scala/com/lookout/borderpatrol/session/SecretsWatcherApi.scala
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  /**
+   * Parse Response from Consul and extract new secret and
+   * current ModifyIndex
+   * @param s
+   * @return
+   */
+  def getSecretInformation(s: String): Option[SecretData] = {
+    implicit def SecretDataCodecJson = casecodec6(SecretData.apply, SecretData.unapply)("CreateIndex", "ModifyIndex", "LockIndex", "Key", "Flags", "Value")
```
trane: i think this implicit def should go outside, i'm not sure why it should be an inner-method for getSecretInformation.
won't you use this json codec for the generator/persister?

borderpatrol-core/src/test/scala/com/lookout/borderpatrol/session/SecretsWatcherApiSpec.scala
wkimeria: Oh, I'll fix this

borderpatrol-core/src/main/scala/com/lookout/borderpatrol/session/SecretsWatcherApi.scala
wkimeria: I will. I'll move it outside

borderpatrol-core/src/main/scala/com/lookout/borderpatrol/session/SecretsWatcherApi.scala
```diff
@@ -0,0 +1,110 @@
...
+  private def decode(secret: String): String = {
+    new String(new sun.misc.BASE64Decoder().decodeBuffer(secret))
+  }
+
+  /**
+   * Handle http response. If 200, return body content, otherwise, return response
+   * @param response
+   * @return
+   */
+  private def handleResponse(response: HttpResponse): Either[HttpResponse, String] = {
```
trane: since you are only caring about the response status, what if you made the return type `Either[Int, String]`, then return `Left(response.getStatus)`?

borderpatrol-core/src/main/scala/com/lookout/borderpatrol/session/SecretStore.scala
```diff
@@ -19,6 +23,48 @@ sealed trait SecretStoreApi {
   def find(f: (Secret) => Boolean): Try[Secret]
 }
 
+case class ConsulSecretStore(watcher: SecretsWatcherApi) extends SecretStoreApi {
+  //During initialization, we want this to be a hard failure that prevents server from starting
+  private[this] var _secrets: Secrets = watcher.initialSecrets match {
```
trane: since this isn't bubbling up an Error value, but rather just throwing an exception. why not make `.initialSecrets` return `Secrets` and throw the exception from `initialSecrets` instead of boxing it in a `Try[Secrets]` and then immediately unboxing it?

borderpatrol-core/src/main/scala/com/lookout/borderpatrol/session/SecretsWatcherApi.scala
```diff
@@ -0,0 +1,110 @@
...
+            secret <- newSecrets(secrets)
+          } yield promise.success(secret)
+        }
+      }
+      println("Retrieved from Consul: " + resp.getContent.toString(Charsets.Utf8))
+    }
+    promise.future
+  }
+
+  private def newSecrets(secrets: SecretData): Option[Secrets] = {
```
trane: one more idea for a more functional approach to this code, move the check for `currentModifyIndex` out of this method and then gate the call to this method with that check. then in the body of the method do:

```
decode(secrets.Value).asSecrets.filter(s => s.ModifyIndex > currentModifyIndex)
```

borderpatrol-core/src/main/scala/com/lookout/borderpatrol/session/SecretsWatcherApi.scala
wkimeria: I was leaving this as response cause I haven't decided how much logging is appropriate (just the response status, or the response body etc)

borderpatrol-core/src/main/scala/com/lookout/borderpatrol/session/SecretsWatcherApi.scala
wkimeria: Changed to 
decode(secrets.Value).asSecrets.filter(s => secrets.ModifyIndex > currentModifyIndex)

since the decoded secret does not have a modifyIndex, only the SecretData does

trane: oh, that won't work then. nm.

PR #14, opened 2015-01-23
Adding tests for Consul Secrets Watcher

wkimeria:


borderpatrol-core/src/main/scala/com/lookout/borderpatrol/session/package.scala
```diff
@@ -161,9 +159,24 @@ package object session {
...
 
+  /**
+   * The SecretData class represents a row of data coming back from the Secret Store back-end (Consul)
+   * The JSON Data structure looks like below (value is Base64 Encoded)
+   * [{"CreateIndex":66,"ModifyIndex":68,"LockIndex":0,"Key":"borderpatrol/currentSecret","Flags":0,"Value":"YmJiYmJi......"}]
+   * THE Value is Base64 encoded, when decoded, it looks like
+   * {"current":{"expiry":{"nanos":100000000000},"id":-29,"entropy":[67,93,65,26,89,123,-88,-59,-82,103,-81,-43,-113,-98,-21,-19]},
+   * "previous":{"expiry":{"nanos":100000000000},"id":-96,"entropy":[-66,5,60,86,12,62,-85,67,72,57,-19,-5,-47,-26,-101,63]}}
+   */
+  case class SecretData(val CreateIndex: Int,
```
trane: why is this in session/package.scala? isn't this a consul only thing?

borderpatrol-core/src/test/scala/com/lookout/borderpatrol/session/SecretsWatcherApiSpec.scala
```diff
@@ -2,38 +2,74 @@ package com.lookout.borderpatrol.session
...
-import com.twitter.finagle.Service
+import com.lookout.borderpatrol.{ConsulService, Response}
 import com.twitter.io.Charsets
-import com.twitter.util.Future
-import com.twitter.util.{Duration, Time}
+import com.twitter.util.{Duration, Future, Time}
 import org.jboss.netty.buffer.ChannelBuffers._
 import org.jboss.netty.handler.codec.http._
 import org.scalatest.{FlatSpec, Matchers}
+import sun.misc.BASE64Encoder
```
trane: i'm using twitter's bijection for all encoding/decoding of base64, i think it is safer for us to do that then to do ad-hoc solutions in different files, you can look at the new InMemoryEncryptedSessionStore for an example on how to use it.

basically, it should look something like this. I would highly recommend doing this in the actual secretswatcher, too.
```
val coder = Injection.connect[String, Array[Byte], Base64String, String]
...
// encode
coder(secrets.asJson)
// decode : Try[String]
coder.invert(base64).get.asSecrets
```

borderpatrol-core/src/main/scala/com/lookout/borderpatrol/session/package.scala
```diff
@@ -161,9 +159,24 @@ package object session {
...
+                        val ModifyIndex: Int,
+                        val LockIndex: Int,
+                        val Key: String,
+                        val Flags: Int,
+                        val Value: String)
+
   implicit def SecretDataCodecJson = casecodec6(SecretData.apply, SecretData.unapply)("CreateIndex", "ModifyIndex", "LockIndex", "Key", "Flags", "Value")
```
trane: i'm currently working on cleaning this package craziness up, maybe i can merge this into what i'm working on and clean this up too.

borderpatrol-core/src/main/scala/com/lookout/borderpatrol/session/package.scala
wkimeria: It is. I'll move it to a different package

borderpatrol-core/src/test/scala/com/lookout/borderpatrol/session/SecretsWatcherApiSpec.scala
wkimeria: Will do

trane:
i'd like to merge this, but it won't let me

wkimeria:
I'll fix this

PR #15, opened 2015-01-23
Cleanup and organize components

trane:
This is a first try at some sane organization for a growing code base.
Hopefully, this will also help us test and see what tests are missing.

### DIFF
--- a/borderpatrol-core/src/main/scala/com/lookout/borderpatrol/Session.scala
+++ b/borderpatrol-core/src/main/scala/com/lookout/borderpatrol/Session.scala
@@ -1,0 +1,51 @@
+package com.lookout.borderpatrol
+
+import com.lookout.borderpatrol.session.id.{SignerComponent, ExpiryComponent, Marshaller, Generator}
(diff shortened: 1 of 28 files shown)


borderpatrol-core/src/main/scala/com/lookout/borderpatrol/Session.scala
```diff
@@ -0,0 +1,51 @@
...
+}
+
+object Session extends SessionFactory with SignerComponent
+                                      with ExpiryComponent
+                                      with SecretStoreComponent
+                                      with SessionStoreComponent {
+
+  val cookieName = "border_session"
+  val entropySize = Constants.SessionId.entropySize
+  implicit val secretStore = InMemorySecretStore(Secrets(Secret(SecretExpiry.currentExpiry), Secret(Time.fromSeconds(100))))
```
wkimeria: Should time for previous secret be 24 hours, and not 100 seconds?

wkimeria:
This looks good.

PR #20, opened 2015-01-30
Add encrypted session memcached store

trane:


trane:
Only one build failed, and it was a dependency issue not the code.

PR #22, opened 2015-04-01
Add pluggable auth

trane:
This includes some house keeping, as well as the first go at a pluggable auth API.


trane:
This is meant to implement #24

Issue #39, opened 2015-07-31
Refactor where Objects and Type Classes are located


trane:
We can clean up the implicit locations and make for a more enjoyable user experience by following some simple patterns.

 - Each public facing entity should have a file named the same thing as the object/trait/type class, e.g. `Session.scala => { trait Session[A]; object Session }`
 - The companion object should be included in the same file
 - Default implicits should be put in that companion object

Example Type-Class: `ToSessionId.scala`

```scala
/**
 * A type class that provides a conversion from some type to a SessionId
 */
trait ToSessionId[A] {
  def apply(a: A): SessionId
}

object ToSessionId {
  /**
   * A convenience method that supports creating a ToSessionId instance from
   * a function.
   */
  def apply[A](converter: A => SessionId): ToSessionId[A] = new ToSessionId[A] {
    def apply(r: A): SessionId = converter(r)
  }

  /**
   * An identity instance for SessionId itself.
   */
  implicit val sessionIdIdentity: ToSessionId[SessionId] = apply(identity)
}
```

PR #55, opened 2015-08-19
Refactor sessionx._

trane:
sessionx was really difficult to understand with the layout and
seemingly random location for different types and functions.

This commit specifically addresses this by presenting a more easily
understood way: a file per concept. Other files were touched to deal
with differing imports.

Further, this commit does away with extraneous implicits and reworks
some of the implicits using a common and modern typeclass pattern

Fixes: #39


core/src/main/scala/com/lookout/borderpatrol/sessionx/Secret.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  private[sessionx] def id: SecretId =
+    EntropyGenerator(idSize).head
+
+  private[sessionx] def entropy: Entropy =
+    EntropyGenerator(entropySize)
+
+  /**
+   * Creates a new [[Secret]] with default expiry of 1 day
+   * @param expiry the time this should expire, defaults to 1 day
```
PeterFCaswell: This is only 1 day because lifetime, above, is 1 day. If someone changes the value of lifetime, the comment is wrong. I'd suggest making the comment say "with default expiry as set above in lifetime", or something like it.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -0,0 +1,46 @@
+package com.lookout.borderpatrol.sessionx
+
+/**
+ * Only two secrets must be in rotation at one time:
```
PeterFCaswell: Does this mean: "Only two secrets may be in rotation at one time." or "Exactly two secrets must be in rotation at all times." or something else?

core/src/main/scala/com/lookout/borderpatrol/sessionx/package.scala
```diff
@@ -24,11 +24,9 @@
...
-import com.twitter.io.Buf
+import argonaut._, Argonaut._
+import com.twitter.finagle.httpx.Cookie
 import com.twitter.util.{Time, Future}
-import com.twitter.finagle.memcachedx
-
-import scala.collection.mutable
 
 /**
  * This crypto introduces types and functions that enable identifying, fetching, and storing web session data. This
```
PeterFCaswell: What do you mean by "This crypto"?

server/src/main/scala/com/lookout/borderpatrol/routers/ServiceMatchers.scala
```diff
@@ -39,7 +39,7 @@ object ServiceMatchers {
      *          Given a host of "sub.subdomain.example.org" and a Set[ServiceIdentifier] of
      *            (ServiceIdentifier("one", "/s", "sub.subdomain"),
      *             ServiceIdentifier("two", "/s2", "sub"))
-   *             return the ServiceIdentifier named "one" because it is the longest matching
+     *          return the ServiceIdentifier named "one" because it is the longest matching
```
PeterFCaswell: Should the name of this object or method (and not just the comment) indicate that it is doing longest longest-prefix match?

trane: I went back and forth with this exact question yesterday while writing it. I came away with the following:
  - Users of the API are expecting the best match for subdomain
  - The longest prefix occurs as only part of this request
  - I have no use case for abstracting out `longestPrefix` to another method, since it is specific to this use case

that being said, it could help readability to break it up, so i will

core/src/main/scala/com/lookout/borderpatrol/sessionx/package.scala
trane: my thoughts were encrypted at the time

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
trane: fixing the ambiguity

core/src/main/scala/com/lookout/borderpatrol/sessionx/Secret.scala
trane: good point, i'll make that explicit

Issue #52, opened 2015-08-10
borderpatrol-server module


trane:
The server module should implement the business logic for routing utilizing the `auth` and `core` modules.

For routing, I think utilizing Finch's `Router[A]` types/functions would be ideal, since we can extend this functionality to support multi-domain default routes as well as path based routing.

For `routers`, I imagine something like this (but that can be based on configuration):

```scala
import com.twitter.util.Future
import io.finch.route.{Extractor, Router}
import shapeless.HNil

object routers {

  private[routers] class SubdomainMatcher(subdomain: String) extends Router[HNil] {
    import Router._
    def apply(input: Input): Option[(Input, () => Future[HNil])] =
      if (input.request.uri.startsWith(subdomain)) Some((input, () => Future.value(HNil)))
      else None

    override def toString: String = s"$subdomain"
  }
 // or ...
   private[routers] case class DefaultServiceDomain(g: String => Option[String])
      extends Extractor[String]("defaultService", identity) {

    import Router._

    override def apply(input: Input): Option[(Input, () => Future[String])] =
      for {
        h <- input.request.host
        s <- g(h)
      } yield (input.drop(1), () => Future.value(s))
  }

  val mappings = Set(("enterprise", "ent"), ("consumer", "my"))
  val supportedDomains = DefaultServiceDomain(s => mappings.find(t => s.startsWith(t._1)).map(_._2))
  val service = string("service") | supportedDomains
}
```

PR #63, opened 2015-08-21
Fix build and update docs

trane:
The build was failing because of missing implicits in server.readers,
this was just an oversight that wasn't caught because we have no tests
in server yet.

Further, server was misnamed as "routers", so this was changed.

Ref: #52

### DIFF
--- a/server/src/main/scala/com/lookout/borderpatrol/server/readers.scala
+++ b/server/src/main/scala/com/lookout/borderpatrol/server/readers.scala
@@ -1,25 +1,26 @@
-package com.lookout.borderpatrol.routers
+package com.lookout.borderpatrol.server
 
+import com.lookout.borderpatrol.sessionx.SecretStoreApi
 import com.twitter.bijection.twitter_util.UtilBijections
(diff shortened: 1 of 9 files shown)


server/src/main/scala/com/lookout/borderpatrol/server/readers.scala
```diff
@@ -1,25 +1,26 @@
-package com.lookout.borderpatrol.routers
+package com.lookout.borderpatrol.server
 
+import com.lookout.borderpatrol.sessionx.SecretStoreApi
 import com.twitter.bijection.twitter_util.UtilBijections
-import com.twitter.finagle.httpx.Request
+import com.twitter.finagle.httpx
```
PeterFCaswell: Should this be "...finagle.httpx._"? With the underscore?

trane: no, i don't want to import `*`. I just wanted to make explicit in the code that it is `httpx.Request`

PR #64, opened 2015-08-26
Update tests and crypto

trane:
This is a collection of 3 commits that properly namespace tests and integrate crypto back into the rest of the design of border patrol.


core/src/main/scala/com/lookout/borderpatrol/crypto/EncryptedSessionStore.scala
```diff
@@ -0,0 +1,71 @@
...
+
+import com.lookout.borderpatrol.sessionx.{EncryptedDataEncoder, SessionDataEncoder, SessionId, Session}
+import com.twitter.io.Buf
+import com.twitter.util.Future
+import com.twitter.finagle.memcachedx
+
+import scala.util.{Success, Failure}
+
+/**
+ * Session store that will store encrypted `Session[_]` data into
```
PeterFCaswell: I'm not sure this sentence parses.

Issue #54, opened 2015-08-18
Add tut as a dependency during builds


trane:
We want our documentation to stay up-to-date -- it's horrible when it doesn't.

Tut will enforce that our documentation examples compile (this includes requiring imports to be scoped properly).

https://github.com/tpolecat/tut

jamescway:
it looks like tut doesn't parse scaladoc and only really parses markdown/html type of files.  Doobie has a tut dir that has all md files as an example.

jamescway:
checking out sbt-doctest

jamescway:
sbt-doctest generates tests based on the scaladocs and runs with the `test` command.  Maybe it would be easier to maintain a set of MD files, @trane do you have any preference?

trane:
we need sbt-doctest regardless.
i would prefer to have great markdown user docs AND scaladoc.

trane:
let's focus on tut and moving our user docs from scaladoc to markdown, and bettering our user docs.

PR #65, opened 2015-08-27
Add tut to parse markdown files

jamescway:
Adds a /tut directory in project root to store markdown files.
Parses MD files and stores output in target/scala-N.NN/tut.

Refs: #2, #54

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -64,6 +64,10 @@ lazy val publishSettings = Seq(
   bintrayOrganization := Some("lookout")
 )
 
+lazy val tutExtraSettings = Seq(
+  tutSourceDirectory := baseDirectory.value / "docs" / "src" / "main" / "tut"
+)
+
 lazy val noPublish = Seq(
   publish := {},
   publishLocal := {}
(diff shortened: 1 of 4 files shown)


jamescway:
fixing the errors in the markdown

build.sbt
```diff
@@ -77,10 +77,16 @@ lazy val docSettings = site.settings ++ ghpages.settings ++ unidocSettings ++ Se
   unidocProjectFilter in (ScalaUnidoc, unidoc) := inAnyProject
 )
 
+lazy val tutExtraSettings = Seq(
+  tutSourceDirectory := baseDirectory.value / "tut"
```
trane: let's follow this conventions set by cats: https://github.com/non/cats/tree/master/docs

`docs/src/{main/tut,site}`

build.sbt
```diff
@@ -77,10 +77,16 @@ lazy val docSettings = site.settings ++ ghpages.settings ++ unidocSettings ++ Se
   unidocProjectFilter in (ScalaUnidoc, unidoc) := inAnyProject
 )
 
+lazy val tutExtraSettings = Seq(
```
trane: take a look at this for integrating with the `docSettings` from above: https://github.com/non/cats/blob/master/build.sbt#L55-L81

build.sbt
jamescway: it seems like when I put the `tutExtraSettings` in `docSettings` or `allSettings` it doesn't get evaluated.  Any idea why?

jamescway:
theres a lot of new boilerplate in the docs, maybe worth trying to silence

trane:
:) nice, it's a great first step.

PR #71, opened 2015-09-21
Added configuration flags and build ServerConfig object

### LINKED ISSUE #70: Add configuration using twitter-util Flags

maheshkelkar:
Using twitter-util flags to accept command line arguments
Using circe to read list of ServiceIdentifier(s) from the JSON file

Fixes #70

### DIFF
--- a/core/src/main/scala/com/lookout/borderpatrol/sessionx/Errors.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/sessionx/Errors.scala
@@ -16,5 +16,3 @@ case class SessionStoreError(msg: String)
 
 case class SecretDecodeError(msg: String)
     extends SessionError(s"An error decoding a Secret occurred: $msg")
-
-class ConfigError(val message: String) extends Exception(message, null)
--- a/core/src/main/scala/com/lookout/borderpatrol/sessionx/Errors.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/sessionx/Errors.scala
@@ -16,3 +16,5 @@ case class SessionStoreError(msg: String)
 
 case class SecretDecodeError(msg: String)
     extends SessionError(s"An error decoding a Secret occurred: $msg")
+
+class ConfigError(val message: String) extends Exception(message, null)
--- a/example/src/main/scala/com/lookout/borderpatrol/example/BorderPatrolApp.scala
+++ b/example/src/main/scala/com/lookout/borderpatrol/example/BorderPatrolApp.scala
@@ -1,5 +1,7 @@
 package com.lookout.borderpatrol.example
 
+import java.net.InetSocketAddress
+
 import com.lookout.borderpatrol.server.ServerConfig
 import com.twitter.finagle.Httpx
 import com.twitter.util.Await
@@ -14,7 +16,8 @@ object BorderPatrolApp extends App {
   val serviceIdsFile = flags("serviceids-file", "bpSids.json",
     "Filename to read Service Identifiers in JSON format")
 
-  val memcachedServers = flags("memcached-servers", ("localhost:1234"),
+  val memcachedServers = flags("memcached-servers",
+    Seq(new InetSocketAddress("localhost", 11211)),
     "Comma separated list of Memcached Server URIs")
 
   val sessionStoreStr = flags("sessionStore", "memory",
@@ -28,7 +31,7 @@ object BorderPatrolApp extends App {
 
   // Instantiate a ServerConfig
   val globalServerConfig = ServerConfig(secretStoreStr(), sessionStoreStr(),
-    memcachedServers().split(",").toList, serviceIdsFile())
+    memcachedServers(), serviceIdsFile())
 
   // Launch the Server
   val server = Httpx.serve(":8080", routes.toService)
--- a/example/src/main/scala/com/lookout/borderpatrol/example/BorderPatrolApp.scala
+++ b/example/src/main/scala/com/lookout/borderpatrol/example/BorderPatrolApp.scala
@@ -13,24 +13,33 @@ object BorderPatrolApp extends App {
 
   val flags = new Flags("Border Patrol Config Flags")
 
-  val serviceIdsFile = flags("serviceids-file", "bpSids.json",
-    "Filename to read Service Identifiers in JSON format")
+  val secretStoreServers = flags("secretStore-servers",
+    Seq[InetSocketAddress](),
+    "Comma separated list of Secret store Server URIs. Overrides the in-memory setting.")
+
+  val inMemorySecretStore = flags("use-InMemorySecretStore", false,
+    "Use in-memory Secret store. ")
+
+  val sessionStoreServers = flags("sessionStore-servers",
+    Seq[InetSocketAddress](),
+    "Comma separated list of Session store Server URIs. Overrides the in-memory setting.")
+
+  val inMemorySessionStore = flags("use-InMemorySessionStore", false,
+    "Use in-memory Session store")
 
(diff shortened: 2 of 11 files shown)


example/src/main/scala/com/lookout/borderpatrol/example/BorderPatrolApp.scala
```diff
@@ -1,11 +1,36 @@
...
+
+object BorderPatrolApp extends App {
   import endpoint._
 
+  val flags = new Flags("Border Patrol Config Flags")
+
+  val serviceIdsFile = flags("serviceids-file", "bpSids.json",
+    "Filename to read Service Identifiers in JSON format")
+
+  val memcachedServers = flags("memcached-servers", ("localhost:1234"),
```
trane: shouldn't this just be `flags("memcached-servers", Seq(new InetSocketAddress("localhost", 11211)))`, that way you can do the following from the cli:

```
run --memcached-servers=10.1.1.1:11211,10.2.1.1:11211,192.168.1.2:11211
```

example/src/main/scala/com/lookout/borderpatrol/example/BorderPatrolApp.scala
```diff
@@ -1,11 +1,36 @@
...
+
+  val secretStoreStr = flags("secretStore", "memory",
+    "Specify type of Secret Store. Options: memory or memcached")
+
+  // Scan args
+  flags.parseOrExit1(args)
+
+  // Instantiate a ServerConfig
+  val globalServerConfig = ServerConfig(secretStoreStr(), sessionStoreStr(),
+    memcachedServers().split(",").toList, serviceIdsFile())
```
trane: the `memcachedServers().split(",").toList` happens automatically and more safely via the implicit behavior of a `Seq[A](???)`.

example/src/main/scala/com/lookout/borderpatrol/example/BorderPatrolApp.scala
maheshkelkar: done.

example/src/main/scala/com/lookout/borderpatrol/example/BorderPatrolApp.scala
maheshkelkar: done. I am using it as Seq[InetSocketAddress]

server/src/main/scala/com/lookout/borderpatrol/server/ServerConfig.scala
```diff
@@ -0,0 +1,70 @@
...
+    sessionStoreStr: String,
+
+    // {"memcachedServer": ["localhost:1234"]}
+    memcachedServers: List[String],
+
+    // [{"name":"one","path":"/customer1","subdomain":"customer1"}]
+    serviceIdsFile: String): ServerConfig = {
+
+    // Secret Store
+    val secretStore = secretStoreStr match {
```
trane: hrm, this is a very interesting (taking the concept of runtime declaration to map to some static instance). i know we discussed this a bit within the context of SSO. i'm not convinced that this is useful here, though.

here is my argument against applying this idea to the `SecretStore` and `SessionStore`:

- When we are configuring a server instance of BP, we already know how we are deploying it
- This is not generically reusable, since users of this would need to kill this method and build their own

Here is an alternative idea:
* Make arguments for `--secretStore-servers=...,...,...` and `--sessionStore-servers=...` which will take `Seq(InetSocketAddress)` that we can pass to the instances that we have chosen, e.g. `MemcachedSessionStore`
* Make another argument for `--use-InMemorySecretStore` and `--use-InMemorySessionStore` to allow for ephemeral/testing

core/src/main/scala/com/lookout/borderpatrol/sessionx/Errors.scala
```diff
@@ -16,3 +16,5 @@ case class SessionStoreError(msg: String)
 
 case class SecretDecodeError(msg: String)
     extends SessionError(s"An error decoding a Secret occurred: $msg")
+
+class ConfigError(val message: String) extends Exception(message, null)
```
trane: this belongs in the main `com.lookout.borderpatrol.server` package (or `.example` depending on what happens with the divergence/convergence of example and server happen)

server/src/test/scala/com/lookout/borderpatrol/server/ServiceMatchersSpec.scala
```diff
@@ -9,7 +10,7 @@ import io.finch.response.Ok
 import io.finch.route._
 import ServiceMatchers._
 
-class ServiceMatchersSpec extends BorderPatrolSuite {
+class ServiceMatchersSpec extends FlatSpec with Matchers with TryValues with OptionValues {
```
trane: this should extend `BorderPatrolSuite`, if this isn't working we should fix what is making it not work.

server/src/main/scala/com/lookout/borderpatrol/server/ServerConfig.scala
```diff
@@ -0,0 +1,72 @@
...
+    // Set of ServiceIdentifier
+    implicit val encodePath: Encoder[Path] = Encoder.instance(p => Json.obj( ("str", Json.string(p.toString())) ))
+    implicit val decodePath: Decoder[Path] = Decoder.instance(c =>
+      for {
+        str <- c.downField("str").as[String]
+      } yield Path(str)
+    )
+
+    val serviceIdentifiers = decode[Set[ServiceIdentifier]](Source.fromFile(serviceIdsFile).mkString) match {
+      case Xor.Right(a) => a.asInstanceOf[Set[ServiceIdentifier]]
```
trane: this unsafe and not in the style/direction that lends itself to composability. let's work through this together and encode this is a more functional style :)

server/src/main/scala/com/lookout/borderpatrol/server/ServerConfig.scala
maheshkelkar: Alternatively, absence of secretStore-servers defaults it to the in-memory secretStore. I don't see any need for another flag for -use-InMemorySecretStore

trane: I agree with you from a developer point of view. But, from an operational perspective this is asking for trouble. Having the potential for unexpected behavior in a cluster of machines in various environments is hard enough to reason about.  Again, this is just an example server - so no big deal. But, in an actual implementation I would be against that default behavior.

maheshkelkar: I have added the flag as well. I wanted to keep it as close as what we are trying to do.

core/src/main/scala/com/lookout/borderpatrol/sessionx/Errors.scala
maheshkelkar: done

server/src/test/scala/com/lookout/borderpatrol/server/ServiceMatchersSpec.scala
maheshkelkar: I haven't gotten a chance to figure out how to include com.lookout.borderpatrol.test (which is in the core/test/.." into "server/test/.."

server/src/main/scala/com/lookout/borderpatrol/server/ServerConfig.scala
maheshkelkar: I haven't changed this yet. Because I want to understand what direction you are suggesting.

maheshkelkar:
No plans to merge this; as these changes are already merged as a part of #80

PR #72, opened 2015-09-23
Define core auth abstractions with docs

trane:
Using the existing auth contracts and implementing the glue for an
example authentication endpoint revealed several issues. To resolve
these, the commit includes:
- fixes to auth related to an actual implementation
- the first circe integration (of more to come)
- finagle-only filters (as opposed to finch Routers in server module)
- brings ServiceMatcher into borderpatrol package
- brings ServiceIdentifier into borderpatrol package
- adds tests for ServiceMatcher


auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -0,0 +1,78 @@
...
+import com.twitter.finagle.httpx.{RequestBuilder, Response, Request}
+import com.twitter.util.Future
+
+object Keymaster {
+  case class Credential(email: String, password: String, serviceId: ServiceIdentifier)
+
+  class IdentifyException extends Exception("")
+  case object AccessDenied extends Exception("no access available for this service")
+
+  case class KeymasterIdentifyReq(credential: Credential) extends IdentifyRequest[Credential]
```
maheshkelkar: How dos this initialize `credential` in the base class i.e. `IdenitifyRequest`? Or is it done auto-magically in Scala?

trane: it doesn't automatically initialize. `KeymasterIdentifyReq` has a constructor that takes a `Credential`. Presumably, the login provider would send the parameters for a `Credential` in a `POST` request to us and we would instantiate it.

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -0,0 +1,78 @@
...
+     */
+    def apply(req: IdentifyRequest[Credential]): Future[IdentifyResponse[Tokens]] =
+      service(api(req.credential)).flatMap(res =>
+        Tokens.derive[Tokens](res.contentString).fold[Future[IdentifyResponse[Tokens]]](
+          err => Future.exception(err),
+          t => Future.value(KeymasterIdentifyRes(t))
+        )
+      )
+  }
+
```
maheshkelkar: I assume the `example` will glue these service/filters as follows:

```
val keymasterService = work-in-progress
val accessService = ServiceFilter(serviceMatchers, serviceIdentifiersSet) andThen KeymasterAccessIssuer(keyMasterService, sessionStore)
val identityService = IdentityFilter andThen KeymasterIdentitiyProvider(accessService, sessionStore)
```

trane: that's the basic idea, yes. :)

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
maheshkelkar: Following 2 declarations made me think that both the base class `IdentifyRequest` and `KeymasterIdentitfyReq` have same variable defined `credential`. Is it defined twice?

```
case class KeymasterIdentifyReq(credential: Credential) extends IdentifyRequest[Credential]
trait IdentifyRequest[A] {
  val credential: A
}
```

trane: no, the `KeymasterIdentifyReq` is an implementation of `IdentifyRequest[Credential]`. `val credential` in the `trait` is instantiated by the `credential: Credential` in the case class constructor.

maheshkelkar:
:+1:

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
maheshkelkar: Thanks

PR #73, opened 2015-10-01
Add LoginFilter for Keymaster

trane:
This login filter will listen to the login service's POST and transform
the request into a IndentifyRequest[Credential] that Keymaster can
understand. Once Keymaster has succesfully provided the identity, that
identity is saved to a new session pointer and the user is redirected to
their original request location with a new cookie

It will also remove the old session, as it is no longer valid

### DIFF
--- a/auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
+++ b/auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
@@ -1,9 +1,12 @@
 package com.lookout.borderpatrol.auth.keymaster
 
+import com.lookout.borderpatrol.util.Combinators.tap
+
+import com.lookout.borderpatrol.sessionx._
 import com.lookout.borderpatrol.{sessionx, ServiceIdentifier}
 import com.lookout.borderpatrol.auth._
-import com.twitter.finagle.Service
-import com.twitter.finagle.httpx.{RequestBuilder, Response, Request}
+import com.twitter.finagle.{Filter, Service}
+import com.twitter.finagle.httpx.{Status, RequestBuilder, Response, Request}
 import com.twitter.util.Future
 
 object Keymaster {
@@ -24,7 +27,7 @@ object Keymaster {
    * identity (master token)
    * @param service Keymaster service
    */
-  case class KeymasterIdentityProvider(service: Service[Request, Response], store: sessionx.SessionStore)
+  case class KeymasterIdentityProvider(service: Service[Request, Response])
       extends IdentityProvider[Credential, Tokens] {
     val endpoint = "/api/auth/service/v1/account_master_token"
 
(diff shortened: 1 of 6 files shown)


auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -47,6 +50,48 @@ object Keymaster {
...
+        maybeReq <- store.get[Request](id)
+      } yield maybeReq.fold(tap(Request("/"))(r => r.host = req.req.host.get))(_.data)
+
+    def apply(req: ServiceRequest,
+              service: Service[IdentifyRequest[Credential], IdentifyResponse[Tokens]]): Future[Response] =
+      createIdentifyReq(req).fold(Future.value(Response(Status.BadRequest)))(credReq =>
+        for {
+          sid <- SessionId.fromRequest(req.req).toFuture
+          originReq <- originalRequestOrDefault(sid, req)
+          tokenResponse <- service(credReq)
```
maheshkelkar: so here we are instantiating the KeymasterIdentityProvider (i.e. sending credentials and waiting for toekn response back

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -47,6 +50,48 @@ object Keymaster {
...
 
   /**
+   * Handles logins to the KeymasterIdentityProvider:
+   * - saves the Tokens after a successful login
+   * - sends the User to their original request location from before they logged in or the default location based on
+   *   their service
+   * @param store
+   * @param secretStoreApi
+   */
+  case class KeymasterLoginFilter(store: sessionx.SessionStore)(implicit secretStoreApi: sessionx.SecretStoreApi)
```
maheshkelkar: I think we will define this as:
val keymasterLoginService = KeymasterLoginFilter andThen KeymasterIdentityProvider

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -47,6 +50,48 @@ object Keymaster {
...
+
+    def apply(req: ServiceRequest,
+              service: Service[IdentifyRequest[Credential], IdentifyResponse[Tokens]]): Future[Response] =
+      createIdentifyReq(req).fold(Future.value(Response(Status.BadRequest)))(credReq =>
+        for {
+          sid <- SessionId.fromRequest(req.req).toFuture
+          originReq <- originalRequestOrDefault(sid, req)
+          tokenResponse <- service(credReq)
+          session <- Session(tokenResponse.identity.id)
+          _ <- store.update[Tokens](session)
```
maheshkelkar: Create a new Session w/ tokens as contents

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -47,6 +50,48 @@ object Keymaster {
...
+          _ <- store.delete(sid)
```
maheshkelkar: Delete old session record. If this operation fails, then how do we cleanup newly created session record (it never made it to the client). Do we have a timeout on the Session record in SessionStore? I am just concerned that we don't leak the records in SessionStore

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -47,6 +50,48 @@ object Keymaster {
...
+        } yield tap(Response(Status.TemporaryRedirect))(res => {
+          res.location = originReq.uri
+          res.addCookie(session.id.asCookie)
```
maheshkelkar: Since we are doing redirect, I assume originReq will be as-is sent back to the user. Right?

core/src/main/scala/com/lookout/borderpatrol/auth/Identity.scala
```diff
@@ -31,7 +31,7 @@ case object EmptyIdentity extends Identity[Nothing]
 case class Id[+A](id: A) extends Identity[A]
 
 object Identity {
-  def apply[A](a: A): Identity[A] =
+  def apply[A](a: A): Id[A] =
     Id(a)
 }
```
maheshkelkar: Sorry, why are we doing this? I am sure its right thing to do.

maheshkelkar:
:+1: Looks good to me. All of my comments are questions for better understanding. If all checks pass, could you please commit them? So that I can add tests for that while you are gone.

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
trane: yup

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
trane: yes, 24 hours by default.

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
trane: no, because it was a `Request` and we need to send a `Response` as a redirect.

core/src/main/scala/com/lookout/borderpatrol/auth/Identity.scala
trane: It's just a constructor for the `Identity` type. Since the non-empty version takes a parameter, it returns the non-empty type -- I was thinking of it like the constructor for `Option[A]` in scala:

```
Option(null) == None
Option(1) == Some(1)
```

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
trane: not "instantiating", we are merely calling the `.apply` method on the `KeymasterIdentityProvider`

PeterFCaswell:
Does it make sense to take questions that need to be answered "for better understanding" and put them into the code in the form of comments? In my experience, these sorts of questions keep coming up again. :-)

PR #75, opened 2015-10-05
Added scalatests for Core and Auth abstractions

### LINKED ISSUE #59: Test coverage should be > 90%

maheshkelkar:
Add scalatests for functionality referenced in #72, #73, #74 (SessionIdFilter, ServiceFilter, IdentityFilter, KeymasterLoginFilter, KeymasterIdentityProvider, KeymasterAccessIssuer)

Fixes #59


maheshkelkar:
@trane I have committed latest set of changes. Please review and let me know you comments

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -42,7 +42,8 @@ object Keymaster {
      * Sends credentials, if authenticated successfully will return a MasterToken otherwise a Future.exception
      */
     def apply(req: IdentifyRequest[Credential]): Future[IdentifyResponse[Tokens]] =
-      service(api(req.credential)).flatMap(res =>
+      client(api(req.credential)).flatMap(res =>
+        //***FIXME: If response is NOT ok, then just return it; dont parse
```
trane: see comments in #borderpatrol about matching on `res.status`

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -73,7 +74,7 @@ object Keymaster {
     def originalRequestOrDefault(id: SessionId, req: ServiceRequest): Future[Request] =
       for {
         maybeReq <- store.get[Request](id)
-      } yield maybeReq.fold(tap(Request("/"))(r => r.host = req.req.host.get))(_.data)
+      } yield maybeReq.fold(throw new SessionStoreError(s"failed for ${id}"))(d => d.data)
```
trane: this is better expressed as, and the name of this method should be changed:

```
      store.get[Request](id).flatMap(_ match {
        case Some(session) => Future.value(session.data)
        case None => Future.exception(SessionStoreError(s"no request stored for $id"))
      })
```

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Tokens.scala
```diff
@@ -91,7 +91,7 @@ object Tokens {
   )
 
   implicit val ServiceTokensEncoder: Encoder[ServiceTokens] = Encoder.instance[ServiceTokens](st =>
-    Json.fromFields(st.services.map(t => (t._1, Json.string(t._2.value))).toSeq)
+    Json.obj(("service_tokens", Json.fromFields(st.services.map(t => (t._1, Json.string(t._2.value))).toSeq)))
```
trane: good catch

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Tokens.scala
```diff
@@ -103,8 +103,12 @@ object Tokens {
       SessionDataEncoder.encodeString.encode(TokensEncoder(tokens).toString())
 
     def decode(buf: Buf): Try[Tokens] =
-      SessionDataEncoder.encodeString.decode(buf).map(Json.string).flatMap(j =>
-        TokensDecoder.decodeJson(j).fold[Try[Tokens]](e => Failure(SessionDataError(e)), t => Success(t)))
+      SessionDataEncoder.encodeString.decode(buf).flatMap(s =>
```
trane: much better

auth/src/test/scala/com/lookout/borderpatrol/auth/keymaster/KeymasterSpec.scala
```diff
@@ -0,0 +1,349 @@
...
+
+    // Execute
+    // Validate
+    val caught = the [java.net.MalformedURLException] thrownBy {
+      val output = new KeymasterAccessIssuer(keymasterTestService, "/api/endpoint", sessionStore)(new KeymasterAccessReq(Id(tokens), one, sessionId))
+    }
+    caught.getMessage should equal ("no protocol: /api/endpoint")
+  }
+
+  it should "Error response from Keymaster service, KeymasterAccessIssuer raises a parsing exception" in {
```
trane: We should be able to consolidate and change how some of this stuff is tested into more "unit" testing.

auth/src/test/scala/com/lookout/borderpatrol/auth/keymaster/KeymasterSpec.scala
```diff
@@ -0,0 +1,349 @@
...
+      def apply(request: Request) = Response(Status.NotAcceptable).toFuture
+    }
+    val sessionId = sessionid.next
+
+    // Execute
+    val output = new KeymasterAccessIssuer(testService, endpoint, sessionStore)(new KeymasterAccessReq(Id(tokens), one, sessionId))
+
+    // Validate
+    //***FIXME: need to revisit this logic
+    val caught = the [ParsingFailure] thrownBy {
```
trane: Do we really care that the `ParsingFailure` had the message `exhausted input`? What happens when the library updates and it no longer sends that error message and instead sends `exhausted all input`? We would have to go through all of this code and update a bunch of tests. But, nothing has broken!

auth/src/test/scala/com/lookout/borderpatrol/auth/keymaster/TokensSpec.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  behavior of "Tokens"
+
+  it should "Verify the Tokens methods" in {
+    tokens.service("service1") should be equals (serviceToken1)
+    tokens.service("service2") should be equals (serviceToken2)
+    tokens.service("service3") should be equals (None)
+  }
+
+  it should "Verify the encode and decode for Tokens" in {
```
trane: we should be testing the behavior/unit of decoding invalid here, instead of up in the filter tests. here is where we care that we get a `Left` instead of a `Right`. forget about exceptions.

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -47,9 +46,9 @@ class BorderFilter[A](store: SessionStore)
  * If the service doesn't exist, it returns a 404 Not Found response
  *
  * @param matchers
- * @param identifiers
+ *
```
trane: why the extra line?

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -66,22 +65,26 @@ case class SessionIdFilter(store: SessionStore)(implicit secretStore: SecretStor
...
-    } yield res) or (for {
-      session <- Session(req.req)
-      _ <- store.update(session)
-    } yield tap(Response(Status.TemporaryRedirect)) { res =>
-        res.location = req.id.login
-        res.addCookie(session.id.asCookie)
-      })
-
+    SessionId.fromRequest(req.req) match {
+      //  Propagate the failures from "service" to the caller
```
trane: you are very focused on "failures". this is the success case, don't worry about calling out "failures" here in a comment.

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -66,22 +65,26 @@ case class SessionIdFilter(store: SessionStore)(implicit secretStore: SecretStor
   extends Filter[ServiceRequest, Response, SessionIdRequest, Response] {
 
   def apply(req: ServiceRequest, service: Service[SessionIdRequest, Response]): Future[Response] =
```
trane: above this line should read something like "Passes the SessionId to the next in the filter chain. If any failures decoding the SessionId occur (expired, not there, etc), we will terminate early and send a redirect". then you can remove the inline comments

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -66,22 +65,26 @@ case class SessionIdFilter(store: SessionStore)(implicit secretStore: SecretStor
...
+          res.addCookie(session.id.asCookie)
+        }
+    }
 }
 
 /**
  * Determines the identity of the requester, if no identity it responds with a redirect to the login page for that
  * service
+ *
+ * Note: this filter does not handle the POST to login url with identity information
```
trane: this is more for you than the user. i think better explanation in the example docs would be more useful than this note. so, in the package.scala docs, we should have how the composition of these filters should work, e.g.:

```
val bpFilter = ServiceFilter andThen SessionIdFilter
val loginFilters = bpFilter andThen ...
val authFilters = bpFilter andThen IdentityFilter(???) andThen AccessFilter(???)
```

core/src/main/scala/com/lookout/borderpatrol/sessionx/SessionId.scala
```diff
@@ -91,12 +91,17 @@ object SessionId {
...
+  def fromCookie(cooki: Option[Cookie])(implicit ev: SecretStoreApi): Try[SessionId] =
+    cooki match {
       case Some(cookie) => SessionId.from[Cookie](cookie)
       case None => Failure(SessionIdError("no border_session cookie"))
     }
 
+  def fromRequest(req: Request)(implicit ev: SecretStoreApi): Try[SessionId] =
+    fromCookie(req.cookies.get("border_session"))
+
+  def fromResponse(rep: Response)(implicit ev: SecretStoreApi): Try[SessionId] =
```
trane: when do we need to get a sessionid from a response?

trane:
@maheshkelkar : i think these tests are a good starting point. i’d like to see some more thought go into making them more about the unit of work as opposed to specifics of underlying implementation.

There is a lot of testing of very specific error cases that come from lower down in the abstraction (where test for those lower-level abstractions should already test). For example, you are testing in multiple places that a very specific parse error occurs with a specific message. But, the thing that calls the parser directly should be testing that, not at the layer of abstraction that we are working in.

We want our "unit" tests to be about the "unit of work" that the thing we are testing is doing. A good rule of thumb is:
- Test the success case (given all the predicates, the computation will result in X)
- Test the specific error cases that should occur at the same level you are testing (is the error in `auth/Errors.scala` and was it specifically "thrown" by the thing I'm testing?).
- Test for general failure given some predicate failure.

I've made the mistake of testing too low-level and it is a costly and messy design decision.

One more tip: If you aren't yet, use the code coverage tools to help you determine if you are missing cases that you should be hitting.

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
maheshkelkar: done

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
maheshkelkar: done

auth/src/test/scala/com/lookout/borderpatrol/auth/keymaster/KeymasterSpec.scala
maheshkelkar: I am not sure what you mean by that. But, lets discuss.

auth/src/test/scala/com/lookout/borderpatrol/auth/keymaster/KeymasterSpec.scala
maheshkelkar: The intent here is not to invoke and test the ParsingFailure, but to ensure that we handle the case where keymaster sends the invalid content.

auth/src/test/scala/com/lookout/borderpatrol/auth/keymaster/TokensSpec.scala
maheshkelkar: Sorry, not sure what you mean?

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
maheshkelkar: fixed

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
maheshkelkar: done

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
maheshkelkar: fixed

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
maheshkelkar: done

core/src/main/scala/com/lookout/borderpatrol/sessionx/SessionId.scala
maheshkelkar: For scalatests

maheshkelkar:
I don't think I have any tests targetting the underlying implementation. As mentioned above, I thought ParsingFailure case was only handling the case where I was ensuring that KeymasterIdentityProvider path was covered.

I have tried to make them all about the "unit". But, elts discuss them during pair-up

auth/src/test/scala/com/lookout/borderpatrol/auth/keymaster/KeymasterSpec.scala
```diff
@@ -0,0 +1,349 @@
...
+    } yield toks
+
+  val keymasterLoginFilterTestService = new Service[IdentifyRequest[Credential], IdentifyResponse[Tokens]] {
+    def apply(request: IdentifyRequest[Credential]) = Future(new KeymasterIdentifyRes(tokens))
+  }
+
+  val keymasterTestService = new Service[Request, Response] {
+    def apply(request: Request) = Response(Status.Ok).toFuture
+  }
+
```
trane:
```
def mkTestService(f: () => Request): Service[Request,Response] = new Service {
  def apply(???).
```

maheshkelkar:
@trane I have update the code as per our discussion. And I think most of missing coverage comes from the "example" and nooks, but we will fix it as we get closer. 

Please review it and let me know your comments.

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -26,28 +24,31 @@ object Keymaster {
...
+      //  Authenticate user by the Keymaster
+      client(api(req.credential)).flatMap(res => res.status match {
+        //  Parse for Tokens if Status.Ok
+        case Status.Ok =>
+          Tokens.derive[Tokens](res.contentString).fold[Future[IdentifyResponse[Tokens]]](
+            err => Future.exception(AccessDenied(Status.InternalServerError, "Failed to parse the Keymaster Identity Response")),
+            t => Future.value(KeymasterIdentifyRes(t))
+          )
+        //  Preserve Response Status code by throwing AccessDenied exceptions
+        case anyOtherStatus => Future.exception(AccessDenied(anyOtherStatus, s"Invalid credentials for user ${req.credential.email}"))
```
trane: this is not quite right, since there is a case where Keymaster is giving a 500 or other 400 error which is not access denied.

This match statement will inevitably be expanded to extend to all of Keymaster's return types, there are many: 401 (unauthorized), 400 (bad request), 403 (forbidden), 500 (keymaster server error), 429 (rate limited)

```
case Status.Unauthorized => ???
case Status.BadRequest => ???
case ... etc
```

Perhaps this is best encapsulated by an exception called `IdentityProviderError` instead of `AccessDenied`.

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -94,34 +96,37 @@ object Keymaster {
...
+            Tokens.derive[Tokens](res.contentString).fold[Future[ServiceToken]](
+              e => Future.exception(AccessDenied(Status.NotAcceptable, "Failed to parse the Keymaster Access Response")),
+              t => t.service(req.serviceId.name).fold[Future[ServiceToken]](
+                Future.exception(AccessDenied(Status.NotAcceptable, s"No access allowed to service ${req.serviceId.name}"))
+              )(st => for {
+                _ <- store.update(Session(req.sessionId, req.identity.id.add(req.serviceId.name, st)))
+              } yield st)
+            )
+          //  Preserve Response Status code by throwing AccessDenied exceptions
+          case anyOtherStatus => Future.exception(AccessDenied(anyOtherStatus, s"No access allowed to service ${req.serviceId.name}"))
```
trane: stylistically use: `case _ => ...` and then in put `AccessDenied(res.status, ...)`

auth/src/test/scala/com/lookout/borderpatrol/auth/keymaster/KeymasterSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+  val one = ServiceIdentifier("one", Path("/ent"), "enterprise", "/a/login")
+  val serviceMatcher = ServiceMatcher(Set(one))
+  val sessionStore = SessionStores.InMemoryStore
+
+  // Request helper
+  def req(subdomain: String, path: String, params: Tuple2[String, String]*): Request =
+    Request(s"http://${subdomain + "."}example.com${path.toString}", params:_*)
+
+  //  Tokens
+  val serviceToken1 = new ServiceToken("SomeServiceTokenData1")
```
trane: fwiw, you don't need to call `new` you can just use `val blah = ServiceToken("...")`, and if fact is not very scala-esque to call the `new` keyword unless you need to.

auth/src/test/scala/com/lookout/borderpatrol/auth/keymaster/KeymasterSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+  val serviceToken2 = new ServiceToken("SomeServiceTokenData2")
+  val serviceTokens = new ServiceTokens().add("service1", serviceToken1)
```
trane: why not cut 6 lines down to 2?

```
val serviceTokens = ServiceTokens(Map("service1" -> ServiceToken("data1"), "service2" -> ServiceToken("data2")))
val tokens = Token(MasterToken("masterT"), serviceTokens)
```

auth/src/test/scala/com/lookout/borderpatrol/auth/keymaster/KeymasterSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+  it should "succeed and return IdentityResponse with tokens received from upstream Keymaster Service" in {
+    val testService = mkTestService[Request, Response] {
+      request => tap(Response(Status.Ok))(res => {
+        res.contentString = TokensEncoder(tokens).toString()
+        res.contentType = "application/json"
+      }).toFuture
+    }
+
+    // Execute
+    val output = new KeymasterIdentityProvider(testService, path)(new KeymasterIdentifyReq(new Credential("foo", "bar", one)))
```
trane: again, don't need to clutter with `new` everywhere.

auth/src/test/scala/com/lookout/borderpatrol/auth/keymaster/KeymasterSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+    val loginRequest = req("enterprise", "/login", ("username" -> "foo"), ("password" -> "bar"))
+
+    // Execute
+    val output = (new KeymasterLoginFilter(sessionStore) andThen keymasterLoginFilterTestService)(new ServiceRequest(loginRequest, one))
+
+    // Validate
+    val caught = the [SessionIdError] thrownBy {
+      Await.result(output)
+    }
+    caught.getMessage should equal ("An error occurred reading SessionId: no border_session cookie")
```
trane: i can't think of a good reason to check this message, what we care about is line #172.

auth/src/test/scala/com/lookout/borderpatrol/auth/keymaster/KeymasterSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+    loginRequest.addCookie(cooki)
+
+    // Execute
+    val output = (new KeymasterLoginFilter(sessionStore) andThen keymasterLoginFilterTestService)(new ServiceRequest(loginRequest, one))
+
+    // Validate
+    val caught = the [SessionStoreError] thrownBy {
+      Await.result(output)
+    }
+    caught.getMessage should be equals ("An error occurred interacting with the session store: failed for SessionId")
```
trane: yeah, again not super helpful to check this message.

auth/src/test/scala/com/lookout/borderpatrol/auth/keymaster/KeymasterSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+    val output = (new KeymasterLoginFilter(sessionStore) andThen keymasterLoginFilterTestService)(new ServiceRequest(loginRequest, one))
+
+    // Validate
+    val caught = the [SessionIdError] thrownBy {
+      Await.result(output)
+    }
+    caught.getMessage should equal ("An error occurred reading SessionId: no border_session cookie")
+  }
+
+  it should "return SessionStoreError if it fails find the original request from sessionStore" in {
```
trane: so, this test helps me to see that maybe we should be encoding the failure to fetch the original request from the session store as a unique error, something like `LoginRedirectError` or `OriginalRequestNotFound` - since this functionality of redirecting to the original request is such a part of the basic design of Border Patrol.

This also allows us to do something interesting with this error message later on, like configuring a default path for times that failures like this happen, e.g. a memcached server dies.

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
maheshkelkar: ok, so your comment only is limited to using IdentityProviderError instead of AccessDenied, right? I will take care of it.

auth/src/test/scala/com/lookout/borderpatrol/auth/keymaster/KeymasterSpec.scala
maheshkelkar: I was using serviceToken1, serviceToken2, etc in the tests below

trane: in that case you can bring it from 6 to 3:

In 3 lines:
```
val (serviceToken1, serviceToken2) = (ServiceToken("data1"), ServiceToken("data2"))
val serviceTokens = ServiceTokens(Map("service1" -> serviceToken1, "service2" -> serviceToken2))
val tokens = Token(MasterToken("masterT"), serviceTokens)
```

In 4 lines:
```
val serviceToken1 = ServiceToken("data1")
val serviceToken2 = ServiceToken("data2"))
val serviceTokens = ServiceTokens(Map("service1" -> serviceToken1, "service2" -> serviceToken2))
val tokens = Token(MasterToken("masterT"), serviceTokens)
```

maheshkelkar:
@trane I have tried to incorporate latest set of comments. Let me know if I have missed anything. Thanks for reviewing and let me know if you have additional comments.

trane:
cool, if travisci passes, we can merge this. :+1:

trane:
looks like there are scalastyle errors, make those updates if you haven't:

```
[error] /home/travis/build/lookout/borderpatrol/auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala:46: File line length exceeds 120 characters
[error] /home/travis/build/lookout/borderpatrol/auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala:50: File line length exceeds 120 characters
[error] /home/travis/build/lookout/borderpatrol/auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala:120: File line length exceeds 120 characters
[error] /home/travis/build/lookout/borderpatrol/auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala:122: File line length exceeds 120 characters
[error] /home/travis/build/lookout/borderpatrol/auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala:128: File line length exceeds 120 characters
```

maheshkelkar:
Yes. had fixed some of them. but missed a couple

Issue #40, opened 2015-08-01
Consul backed Secret Store


trane:
We currently have an `InMemorySecretStore` that does well for mocking, but we need to store `Secrets` somewhere that provides a consistent view of the current state of `Secrets`.

Obviously, Zookeeper is a great place to store these things, but since we have a use-case for Consul as well.

All that needs to be done is to provide a Consul backend for a `SecretStoreApi` interface, making sure it is as purely functional as you can -- why not have the entire state of everything ever forever?

trane:
So I think of this as a few abstractions:
 * a way to store into consul (serialize to json) external to individual server processes
 * a way to cache secrets in-memory to avoid fetching from consul constantly
 * a way to poll/watch consul for new secrets

PeterFCaswell:
Why the cache? Fetching from consul should be a fast as memcache.
I can understand we might want a caching mechanism at the SecretStore level, but for the InMemorySecretStore and the ConsulSecretStore, it should be a no-op. I would think.

PeterFCaswell:
Here are more thoughts on this.

I think I'd consider the abstract SecretStore to be comprised of (forgive the Javaish terminology):
  1. Medium Interface
  2. Serializer
  3. Caching Policy
  4. Polling Policy (which also needs the Medium Interface).

The medium interface is straightforward, or should be: a way to read/write/create/delete, whatever.

The serializer is likewise straightforward. Though I wonder if it needs an interface for serializing the key as well as the data.

The caching policy is something that could be generic across multiple objects. Especially we don't expect to read and write through the cache (and are willing to query cache, then medium). For through-caches, this'll need to have the medium interface. The simplest cache is obviously no cache. And there are no end to caching strategies. They can even be stacked.

The Polling policy needs the medium interface, but it can also be a no-op if that's what one wants.

Genericizng the policies gives you some hope of reusing them outside the project, or even of pulling in other code for the policies. It also gives you the chance to change them based on configuration parameters rather than code. And that would hopefully alleviate your concerns about spof or throughput on your storage medium.

trane:
@PeterFCaswell yes, sounds like we are looking at this exactly the same way.

There are two existing interfaces already defined in bp:

`SecretStoreApi` which is the only interface for clients to retrieve the `current` and `previous` rotating secrets, see: `sessionx.SecretStoreApi`.

also, serializing is defined in the `Encoder` type class (`SecretEncoder`) - and reusable.

trane:
[...]

PR #76, opened 2015-10-09
Consul secretapi

tejom:
For issue #40


core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -46,5 +60,213 @@ object SecretStores {
...
+  /**
+  *A store to access the current and previous [[com.lookout.borderpatrol.sessionx.Secret]] stored in the consul server.
+  * It will use a in memory cache before making a request to the server
+  *
+  *@param consul An instance of [[com.lookout.borderpatrol.sessionx.ConsulConnection]] to make requests to the
+  *consul server
+  *@param poll How often in seconds to check for updated Secrets on the consul server
+  **/
+  class ConsulSecretStore(consul: ConsulConnection ,poll: Int) extends SecretStoreApi {
+    val cache = new ConsulSecretCache(poll,consul)
```
trane: what if we used a infinite `Stream` instead?

Something like:
```
val stream: Stream[Secrets] = Stream.continually(consul.secrets)

...
// Then we can define current/previous like this
def current: Secret = stream.head.current
def previous: Secret = stream.head.previous

// this means we need to handle program failure in the class we've implemented for val "consul"
```

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -46,5 +60,213 @@ object SecretStores {
...
+  }
+  /**
+  *Stores the secrets stored on the consul server.
+  *Polls the consul server and updates an inmemory cache for SecretStore
+  *Uses a [[scala.collection.mutable.HashMap]]
+  *
+  *@param poll How often in seconds to check for updates
+  *@param consul An instance on ConsulConnection to make connections with the consul server
+  **/
+  class ConsulSecretCache(poll: Int, consul: ConsulConnection) extends Runnable  {
```
trane: what would this class look like if it's only api was a function called `def secrets: Secrets`?

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -46,5 +60,213 @@ object SecretStores {
...
+    }
+
+  }
+  /**
+  *class to interface with consul kv store
+  *
+  *@param consul Finagle server to send and get requests to the server
+  *@param host host name of the consul server to connect to ex: "localhost"
+  **/
+  class ConsulConnection(consul: Service[httpx.Request, httpx.Response],host: String) {
```
trane: We can work on this class a bit to make it more idiomatic, but this is a great start.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -46,5 +60,213 @@ object SecretStores {
...
+    def apply(consulUrl: String, consulPort: String,poll: Int):ConsulSecretStore = {
+      val apiUrl = s"$consulUrl:$consulPort"
+      val client: Service[httpx.Request, httpx.Response] = Httpx.newService(apiUrl)
+      val consulConnection = new ConsulConnection(client,consulUrl)
+
+      val c = new ConsulSecretStore(consulConnection,poll)
+      c
+    }
+    /**
+    *Argonaut lets you use the response from consul as a class to get attributes from.
```
trane: make sure you merge with `master`, i think we replaced the `argonaut` dependency with a `circe` one.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -1,5 +1,19 @@
 package com.lookout.borderpatrol.sessionx
 
+import org.jboss.netty.handler.codec.http.{DefaultHttpRequest, HttpRequest, HttpResponse, HttpVersion, HttpMethod}
```
trane: we can get away without talking to the underlying netty libraries. we want to use our nice wrappers from finagle.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
trane: looks like we still have argonaut, nm

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -46,5 +58,197 @@ object SecretStores {
...
+    private def rotateSecret: Unit={
+      val s = cacheBuffer.last.current
+      val n = newBuffer.last
+      val newSecrets = Secrets(n,s)
+      cacheBuffer.append(newSecrets)
+    }
+    /**
+    *Get the secret at current from the consul server or returns None
+    **/
+    private def pollCurrent: Future[Option[Secret]] = {
```
trane: Having two separate endpoints for `current` and `previous` is a dangerous pattern. 

First: semantically it's misleading because there are ever 2 valid `Secret`s at a time, there is never a case where 1 secret is valid. This is why there is a `Secrets` model.

Second: by separating these endpoints you are effectively allowing for the case that we could have inconsistency where you can have: `current == previous`, `<remote current> != current`, `<remote previous> != previous`, `current == <remote previous>`. None of these are valid.

Instead, we should poll for `Secrets` which only allow the following possibilities: `Secrets == <remote Secrets>` or `Secrets != <remote Secrets>`

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -46,5 +58,197 @@ object SecretStores {
...
+  /**
+  *Stores the secrets stored on the consul server.
+  *Polls the consul server and keeps an inmemory cache of Secrets
+  *
+  *@param poll How often in seconds to check for updates
+  *@param consul An instance on ConsulConnection to make connections with the consul server
+  **/
+  case class ConsulSecretCache(poll: Int, consul: ConsulConnection) extends Runnable  {
+    val cacheBuffer= collection.mutable.ArrayBuffer[Secrets]( Secrets(Secret(),Secret()) )
+    var newBuffer = collection.mutable.ArrayBuffer[Secret]()
```
trane: this now becomes `ArrayBuffer[Secrets]`

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -46,5 +58,197 @@ object SecretStores {
...
+  }
+  /**
+  *Stores the secrets stored on the consul server.
+  *Polls the consul server and keeps an inmemory cache of Secrets
+  *
+  *@param poll How often in seconds to check for updates
+  *@param consul An instance on ConsulConnection to make connections with the consul server
+  **/
+  case class ConsulSecretCache(poll: Int, consul: ConsulConnection) extends Runnable  {
+    val cacheBuffer= collection.mutable.ArrayBuffer[Secrets]( Secrets(Secret(),Secret()) )
```
trane: this is funny... you are using a `mutable.ArrayBuffer` instead of a `immutable.Stream`. almost the same thing :)

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -46,5 +58,197 @@ object SecretStores {
...
+      while(true){
+        for {
+          n <- pollCurrent
+        } yield ( n match {
+            case Some(s) => newBuffer.append(s)
+          })
+        Thread.sleep( poll * 1000)
+      }
+    }
+    private def rotateSecret: Unit={
```
trane: there is a race condition here

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -46,5 +58,197 @@ object SecretStores {
...
+    **/
+    def secrets: Secrets ={
+      val s = cacheBuffer.last
+      if(s.current.expired && newBuffer.last != s.current) {rotateSecret; cacheBuffer.last}
+      else s
+    }
+    /**
+    *Checks if the secret is knows. Rotates if the secret is the new secret.
+    **/
+    def find(f: Secret=>Boolean): Option[Secret] = {
```
trane: race condition in here

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -46,5 +58,197 @@ object SecretStores {
...
+  *@param consul An instance on ConsulConnection to make connections with the consul server
+  **/
+  case class ConsulSecretCache(poll: Int, consul: ConsulConnection) extends Runnable  {
+    val cacheBuffer= collection.mutable.ArrayBuffer[Secrets]( Secrets(Secret(),Secret()) )
+    var newBuffer = collection.mutable.ArrayBuffer[Secret]()
+    /**
+    *Checks if the current secret is expired and if there is a new secret available and rotates the secrets.
+    *Then returns the latest secret
+    **/
+    def secrets: Secrets ={
```
trane: race condition in here, too

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
tejom: I was going to ask you about this. I was playing around with using stream. The only way I was able to get to work was using append and doing something like `var a  Stream[A] ` and doing `a = a :+ secret` It seemed forced doing it that way. Array lets you use the functions which seems a little more readable.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
tejom: Switching the new buffer to from [Secret] to [Secrets]  should take of this.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
tejom: These should be taken care of with the `this.synchronized` . Some of the alternatives for concurrency seem like overkill like bringing in akka actors. Rotate is the only method that will write to the array so the block should prevent anything else from writing to it.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -46,5 +58,197 @@ object SecretStores {
...
+    }
+    /**
+    *Get just the decoded value for a key from consul as Future[String]. To get the full json response from Consul
+    *use getConsulRepsonse
+    *
+    *@param k the key to get the value for
+    **/
+    def getValue[A,B](k: String): Future[Try[String]] = {
+      val s = getConsulResponse("/v1/kv/" + k)
+      val response = s.map( a => a.decodeOption[List[ConsulSecretStore.ConsulResponse]].getOrElse(Nil) )
```
tejom: I was working on trying to get rid of the List here. decodeOption returns a List of the class ConsulResponse. 
The other way to do this would be similar to how it decoding is done with secret and secrets. For some reason that isn't working. It will parse to a Json but that Json wont become a ConsulResponse and the error given doesnt give any reason why it fails.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -45,6 +55,213 @@ object SecretStores {
...
+    *
+    *@param k key to get a consul json response from
+    **/
+    private def consulResponse(k: String): Future[Buf] = {
+      val req = httpx.Request(httpx.Method.Get, k)
+      req.host = host
+      consul(req).map(a => a.content)
+    }
+
+    private def base64ToString(s: String): Option[String] ={
```
tejom: I ended up putting this function back. The value in the response is a base64 encoded String. twitter.util.Base64StringEncoder returns an Array[Byte] that needs to be converted to a string

core/src/main/scala/com/lookout/borderpatrol/sessionx/Encoder.scala
```diff
@@ -199,7 +202,34 @@ object SecretEncoder {
...
+    implicit val SecretsCodecJson: argonaut.CodecJson[Secrets] =
+      casecodec2(Secrets.apply, Secrets.unapply)("current","previous")
+
+    def encode(s: Secrets): Json={
+      s.asJson
+    }
+
+    def decode(json: Json): Try[Secrets]= {
+      json.as[Secrets].toDisjunction.fold[Try[Secrets]](
+        e => Failure( new Throwable(e._1)),
```
trane: this needs to be something other than `Throwable`, use/create a type that describes what the failure is. read the code above this for examples.

core/src/main/scala/com/lookout/borderpatrol/sessionx/Encoder.scala
```diff
@@ -199,7 +202,34 @@ object SecretEncoder {
...
+object SecretsEncoder {
+
+  implicit object EncodeJson extends SecretsEncoder[Json] {
+    import SecretEncoder.EncodeJson._ //this is neccesary for it to compile otherwise it complains
+    //about not knowing about the implicit object
+
+    implicit val SecretsCodecJson: argonaut.CodecJson[Secrets] =
+      casecodec2(Secrets.apply, Secrets.unapply)("current","previous")
+
+    def encode(s: Secrets): Json={
```
trane: no need for braces

core/src/main/scala/com/lookout/borderpatrol/sessionx/Encoder.scala
```diff
@@ -199,7 +202,34 @@ object SecretEncoder {
...
+      s.asJson
+    }
+
+    def decode(json: Json): Try[Secrets]= {
```
trane: whitespace issue with `=`

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -45,6 +55,213 @@ object SecretStores {
...
+  *@param poll How often in seconds to check for updated Secrets on the consul server
+  **/
+  case class ConsulSecretStore(consul: ConsulConnection ,poll: Int) extends SecretStoreApi {
+    val cache = ConsulSecretCache(poll,consul)
+    new Thread( cache ).start
+
+    /**
+    *Get the current secret from the cache layer
+    **/
+     def current: Secret = {
```
trane: no need for braces

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -45,6 +55,213 @@ object SecretStores {
...
+  }
+  /**
+  *A store to access the current and previous [[com.lookout.borderpatrol.sessionx.Secret]] stored in the consul server.
+  * It will use a in memory cache before making a request to the server
+  *
+  *@param consul An instance of [[com.lookout.borderpatrol.sessionx.SecretStores.ConsulConnection]] to make requests
+  *to the consul server
+  *@param poll How often in seconds to check for updated Secrets on the consul server
+  **/
+  case class ConsulSecretStore(consul: ConsulConnection ,poll: Int) extends SecretStoreApi {
```
trane: whitespace with `,`

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -45,6 +55,213 @@ object SecretStores {
...
+    val cache = ConsulSecretCache(poll,consul)
```
trane: whitespace with `,`

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -45,6 +55,213 @@ object SecretStores {
...
+    *Get the current secret from the cache layer
+    **/
+     def current: Secret = {
+      cache.secrets.current
+    }
+
+    /**
+    *Get the previous secret from the cache layer
+    **/
+    def previous: Secret = {
```
trane: no need for braces

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -45,6 +55,213 @@ object SecretStores {
...
+      cache.secrets.previous
+    }
+
+    /**
+    *Look for the Secret being checked in the function. Checks if this exists in the cache layer
+    *If it is found only existing in consul and not in current or previous a cache rotation will be triggered
+    *Returning None suggests the servers are extremely out of sync with each other or the connection to consul has
+    *failed
+    **/
+    def find(f: Secret => Boolean): Option[Secret] = {
```
trane: no need for braces

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
trane: whitespace issues with `,` and `)`

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -45,6 +55,213 @@ object SecretStores {
...
+  case class ConsulSecretCache(poll: Int, consul: ConsulConnection) extends Runnable  {
+    val cacheBuffer= collection.mutable.ArrayBuffer[Secrets]( Secrets(Secret(),Secret()) )
+    val newBuffer = collection.mutable.ArrayBuffer[Secrets]()
+    /**
+    *Checks if the current secret is expired and if there is a new secret available and rotates the secrets.
+    *Then returns the latest secret
+    **/
+    def secrets: Secrets = {
+      val s = cacheBuffer.last
+      if(s.current.expired) rotateSecret
```
trane: whitespace with `if(`

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -45,6 +55,213 @@ object SecretStores {
...
+      else s
+    }
+
+    /**
+    *Checks if the secret is knows. Rotates if the secret is the new secret.
+    **/
+    def find(f: Secret=>Boolean): Option[Secret] = {
+      val lastSecrets = cacheBuffer.last
+
+      if( f(lastSecrets.current)) Some(lastSecrets.current)
```
trane: whitespace issue with `if( (...)` => `if ((...))`

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -45,6 +55,213 @@ object SecretStores {
...
+      else if ( f(lastSecrets.previous)) Some(lastSecrets.previous)
+      else if ( f(rotateSecret.current)) Some(cacheBuffer.last.current)
+      else None
+    }
+
+    /**
+    *Continously poll the consul server at the interval passed to the Class when it was created
+    *updates the store for a possibly new Secret to be used
+    **/
+    def run(): Unit = {
```
trane: no need for braces

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -45,6 +55,213 @@ object SecretStores {
...
+      while(true){
```
trane: whitespace issues

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -45,6 +55,213 @@ object SecretStores {
...
+        for {
+          n <- pollSecrets
+        } yield  n match {
+            case Some(s) => newBuffer.append(s)
+            case None => ()
+          }
+        Thread.sleep( poll * 1000)
```
trane: whitespace `( ...)` => `(...)`

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -45,6 +55,213 @@ object SecretStores {
...
+      }
+    }
+
+    private def needsRotation: Boolean =
+      newBuffer.last.current != cacheBuffer.last.current
+
+    private def rotateSecret: Secrets = {
+      if (needsRotation){
```
trane: whitespace `){` => `) {`

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -45,6 +55,213 @@ object SecretStores {
...
+        this.synchronized {
+          if (needsRotation)
+            cacheBuffer.append(newBuffer.last)
+          }
```
trane: what is this `}` closing?

tejom: This closes the this.synrchonized block

core/src/test/scala/com/lookout/borderpatrol/test/sessionx/SecretStoresSpec.scala
```diff
@@ -24,4 +26,38 @@ class SecretStoresSpec extends BorderPatrolSuite {
     store.find(_.id == current.id).value shouldBe current
     store.find(_.id == invalid.id) shouldBe None
   }
+
+  val secretsJsonString = SecretsEncoder.EncodeJson.encode(helpers.secrets.secrets).nospaces
```
trane: all of these tests belong in their own "behavior", because this is specifically the consul secret store.

core/src/test/scala/com/lookout/borderpatrol/test/sessionx/SecretStoresSpec.scala
```diff
@@ -24,4 +26,38 @@ class SecretStoresSpec extends BorderPatrolSuite {
...
+  it should "always return a secret" in {
+    val consulConnection = helpers.MockConsulClient
+    consulConnection.set(ConsulSecretsKey, secretsJsonString)
+    val c = new ConsulSecretStore(consulConnection, 10, helpers.secrets.secrets)
+    c.current should not be null
+    c.previous should not be null
+  }
+
+  it should "always return a secret when there is a connection problem" in {
```
trane: this seems more of a test of the caching layer, no?

core/src/test/scala/com/lookout/borderpatrol/test/sessionx/SecretStoresSpec.scala
```diff
@@ -24,4 +26,38 @@ class SecretStoresSpec extends BorderPatrolSuite {
...
+    val consulConnection = helpers.MockConsulClient
+    consulConnection.set(ConsulSecretsKey, "{'invalid':'json'}")
+    val c = new ConsulSecretStore(consulConnection, 10, helpers.secrets.secrets)
+    c.current shouldBe current
+    c.previous shouldBe previous
+  }
+
+  it should "rotate and return the current secret after finding a new current secret" in {
```
trane: this also seems like it's part of a different spec for the caching layer.

core/src/test/scala/com/lookout/borderpatrol/test/sessionx/SecretStoresSpec.scala
```diff
@@ -24,4 +26,38 @@ class SecretStoresSpec extends BorderPatrolSuite {
...
+    val newConsulConnection = helpers.MockConsulClient
+    newConsulConnection.set(ConsulSecretsKey, secretsJsonString)
+    val c = new ConsulSecretStore(newConsulConnection, 10, Secrets(previous, previous))
+    Thread.sleep(1000) //allows the polling function to start appending
```
trane: this will fail randomly. how can you set up this test to ensure that no matter how long it takes to begin appending, that once it has appended the test will complete?

core/src/test/scala/com/lookout/borderpatrol/test/sessionx/SecretStoresSpec.scala
```diff
@@ -24,4 +26,38 @@ class SecretStoresSpec extends BorderPatrolSuite {
...
+    c.find(_.id == current.id)
+    c.current shouldBe current
+  }
+
+  it should "rotate when the secret is expired" in {
+    val newConsulConnection = helpers.MockConsulClient
+    newConsulConnection.set(ConsulSecretsKey, secretsJsonString)
+    val c = new ConsulSecretStore(newConsulConnection, 10, Secrets(previous, previous))
+    Thread.sleep(1000) //allows the polling function to start appending
```
trane: this will fail randomly, too. same as above.

core/src/test/scala/com/lookout/borderpatrol/test/sessionx/helpers.scala
```diff
@@ -36,4 +39,17 @@ object helpers {
...
        Session(sessionid.next, a)
    }
+
+  object MockConsulClient extends ConsulClient {
+    val mockRemote = collection.mutable.HashMap.empty[String, String]
+
+    override def value(key: String): Future[Try[String]] =
+      Future.value(Success(mockRemote(key)))
+
+    override def set(k: String, v: String): Future[httpx.Response] = {
```
trane: cool idea.

core/src/test/scala/com/lookout/borderpatrol/test/sessionx/helpers.scala
```diff
@@ -1,7 +1,10 @@
 package com.lookout.borderpatrol.test.sessionx
 
-import com.lookout.borderpatrol.sessionx.SecretStores.InMemorySecretStore
-import com.twitter.util.{Await, Time}
+import com.lookout.borderpatrol.sessionx.SecretStores.{ConsulClient, ConsulConnection, InMemorySecretStore}
```
trane: let's import these within an object in the helpers

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -45,6 +56,239 @@ object SecretStores {
...
+    /**
+     * Set the given key to the given Value. Both are strings
+     *
+     * @param k the key to set
+     * @param v the value of the key
+     */
+    def set(k: String, v: String): Future[httpx.Response] = {
+      val currentData = Buf.Utf8(v)
+      val update: httpx.Request = httpx.RequestBuilder()
+        .url(s"http://$host:$port/v1/kv/$k")
```
trane: should the host be already handled by the `consul` service? shouldn't we be able to just issue a `Put /v1/kv/$k` without putting the prefix of `http://$host:$port` there?

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -45,6 +56,239 @@ object SecretStores {
...
+     */
+    def value(key: String): Future[Try[String]] =
+      consulResponse("/v1/kv/" + key) map { buf =>
+        (for {
+          body <- Buf.Utf8.unapply(buf)
+          rlist <- body.decodeOption[List[ConsulResponse]]
+          res <- rlist.headOption
+          value <- base64ToString(res.value)
+        } yield value) match {
+          case None => Failure(new Exception("Unable to decode consul response"))
```
trane: let's make this a consul specific exception.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -45,6 +56,239 @@ object SecretStores {
...
+  }
+
+  /**
+   * class to interface with consul kv store
+   *
+   * @param consul Finagle server to send and get requests to the server
+   * @param host host name of the consul server to connect to ex: "localhost"
+   */
+  class ConsulConnection(consul: Service[httpx.Request, httpx.Response], host: String, port: String)
+    extends ConsulClient {
```
trane: this should be 4 spaces in from the start of `class`:

```
class ...
    extends
```

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -45,6 +56,239 @@ object SecretStores {
...
+        case Failure(e) => None
+      })
+
+    /**
+     * Returns a Try[Secret] from json as a [String]
+     *
+     * @param s A json string with information to create a Secret
+     */
+    private def secretsTryFromString(s: String): Try[Secrets] = {
+      import scalaz._
```
trane: do we need to `import` every time? why not just once in this object?

trane:
@tejom this consul secret store is large enough now, i think it deserves to be in its own namespace'd folder.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -45,6 +56,239 @@ object SecretStores {
...
+        case -\/(e) => util.Failure(new Exception(s"Failed with: $e"))
+        case \/-(j) => SecretsEncoder.EncodeJson.decode(j)
+      }
+    }
+  }
+
+  /**
+   * Define functions for a ConsulClient connection. Used by the the ConsulConnection and MockConsulClient
+   */
+  trait ConsulClient {
```
trane: why make this trait? there is only one instance of this, `ConsulConnection`. the mock client can just extend the consulconnection class.

tejom: I tried object `MockConsulClient extends ConsulConnection ` at first but I got an error about unspecified value parameters.

trane:
how is this going @tejom?

PR #79, opened 2015-10-19
Fix borderpatrol example module and weave core and auth module together

### LINKED ISSUE #78: Fix borderpatrol example module
The example module should weave core and auth modules along with the business logic.

maheshkelkar:
Fixes #78


maheshkelkar:
@trane please note that coverage went up, as I excluded the "example" from the code coverage

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -28,7 +27,7 @@ object Keymaster {
    * @param path Keymaster service endpoint path
    */
   case class KeymasterIdentityProvider(client: Service[Request, Response], path: Path)
-      extends IdentityProvider[Credential, Tokens] {
+    extends IdentityProvider[Credential, Tokens] {
```
trane: please keep this 4 spaces in from where the `case class` definition is. it's confusing to have `extends` at the same position as `def`s

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -57,38 +56,37 @@ object Keymaster {
...
    * - saves the Tokens after a successful login
    * - sends the User to their original request location from before they logged in or the default location based on
-   *   their service
+   * their service
    * @param store
    * @param secretStoreApi
    */
   case class KeymasterLoginFilter(store: SessionStore)(implicit secretStoreApi: SecretStoreApi)
-      extends Filter[ServiceRequest, Response, IdentifyRequest[Credential], IdentifyResponse[Tokens]] {
+    extends Filter[SessionIdRequest, Response, IdentifyRequest[Credential], IdentifyResponse[Tokens]] {
```
trane: here, too

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -57,38 +56,37 @@ object Keymaster {
...
 
-    def createIdentifyReq(req: ServiceRequest): Option[IdentifyRequest[Credential]] =
+    def createIdentifyReq(req: SessionIdRequest): Option[IdentifyRequest[Credential]] =
       for {
-        u <- req.req.params.get("username")
-        p <- req.req.params.get("password")
-      } yield KeymasterIdentifyReq(Credential(u, p, req.id))
+        u <- req.req.req.params.get("username")
```
trane: wow, that `req.req.req` would go away if i'd modeled this better. sorry.

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -103,7 +101,7 @@ object Keymaster {
    * @param store Session store
    */
   case class KeymasterAccessIssuer(client: Service[Request, Response], path: Path, store: SessionStore)
-      extends AccessIssuer[Tokens, ServiceToken] {
+    extends AccessIssuer[Tokens, ServiceToken] {
```
trane: here, too

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -135,4 +133,17 @@ object Keymaster {
       )(t => Future.value(t)).map(t => KeymasterAccessRes(Access(t)))
   }
 
+  /**
+   * This filter acquires the access and then forwards the request to upstream service
+   *
+   * @param upstreamClient
+   */
+  case class KeymasterAccessFilter(upstreamClient: Service[Request, Response])
+    extends Filter[AccessIdRequest[Tokens], Response, AccessRequest[Tokens], AccessResponse[ServiceToken]] {
```
trane: here, too

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -135,4 +133,17 @@ object Keymaster {
...
+    def apply(req: AccessIdRequest[Tokens],
+              accessService: Service[AccessRequest[Tokens], AccessResponse[ServiceToken]]): Future[Response] =
+      accessService(AccessRequest(req.id, req.req.req.serviceId, req.req.sid)).flatMap(accessResp =>
```
trane: we should definitely clean this `req.req.req.req` and `access.access` stuff up soon. not for this pr.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SessionStore.scala
```diff
@@ -90,9 +90,11 @@ object SessionStores {
         case _ => None.toFuture
       }
 
-    def update[A](session: Session[A])(implicit ev: SessionDataEncoder[A]): Future[Unit] =
+    def update[A](session: Session[A])(implicit ev: SessionDataEncoder[A]): Future[Unit] = {
+      delete(session.id)
```
trane: interesting, why do we need to `delete`? i bet if we set line L95 to `store.set`

maheshkelkar: For InMemoryStore, I don't see any "set" method or a overloaded addOrUpdate method.

maheshkelkar:
@trane I have taken care of all comments, except delete/store.set

PR #82, opened 2015-11-02
Use json for entire config and validate the configuration

maheshkelkar:
Also:
- rename identityProvider, loginProvider, accessProvider, etc to identityManager,
  loginManager, accessManager, etc
- Move keymasterIdentityProvider service chain and keymasterAccessIssuer service
  chain to keymaster (common) code.
- Add Binders to dynamically bind to upstream or identity or access or login
  service on the fly.
- Modify ServiceIdentifier to include loginManager and hosts.
- Modify ServiceMatchers to match subdomain first and then match the
  serviceIdentifier path or loginManager paths.

### DIFF
--- a/auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
+++ b/auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
@@ -2,7 +2,8 @@ package com.lookout.borderpatrol.auth.keymaster
 
 import com.lookout.borderpatrol.util.Combinators.tap
 import com.lookout.borderpatrol.sessionx._
-import com.lookout.borderpatrol.ServiceIdentifier
+import com.lookout.borderpatrol._
+import com.lookout.borderpatrol.Binder._
 import com.lookout.borderpatrol.auth._
 import com.twitter.finagle.httpx.path.Path
 import com.twitter.finagle.{Filter, Service}
@@ -23,22 +24,25 @@ object Keymaster {
   /**
    * The identity provider for Keymaster, will connect to the remote keymaster server to authenticate and get an
    * identity (master token)
-   * @param client Keymaster service
-   * @param path Keymaster service endpoint path
+   * @param binder Binder to bind to Keymaster identity service
    */
-  case class KeymasterIdentityProvider(client: Service[Request, Response], path: Path)
+  case class KeymasterIdentityProvider(binder: ManagerBinder)
       extends IdentityProvider[Credential, Tokens] {
 
     def api(cred: Credential): Request =
-      Request(Method.Post,
-        Request.queryString(path.toString, ("e", cred.email), ("p", cred.password), ("s", cred.serviceId.name)))
+      tap(Request(Method.Post,  cred.serviceId.loginManager.identityManager.path.toString))(req => {
+        req.contentType = "application/x-www-form-urlencoded"
+        req.contentString = Request.queryString(("e", cred.email), ("p", cred.password), ("s", cred.serviceId.name))
+          .drop(1) /* Drop '?' */
+      })
 
     /**
      * Sends credentials, if authenticated successfully will return a MasterToken otherwise a Future.exception
      */
     def apply(req: IdentifyRequest[Credential]): Future[IdentifyResponse[Tokens]] =
       //  Authenticate user by the Keymaster
-      client(api(req.credential)).flatMap(res => res.status match {
+      binder(BindRequest(req.credential.serviceId.loginManager.identityManager, api(req.credential)))
+        .flatMap(res => res.status match {
         //  Parse for Tokens if Status.Ok
         case Status.Ok =>
           Tokens.derive[Tokens](res.contentString).fold[Future[IdentifyResponse[Tokens]]](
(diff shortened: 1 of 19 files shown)


core/src/main/scala/com/lookout/borderpatrol/Binder.scala
```diff
@@ -0,0 +1,67 @@
...
+
+import com.twitter.finagle.{Httpx, Service}
+import com.twitter.finagle.httpx.{Response, Request}
+import com.twitter.util.Future
+import scala.collection.mutable
+
+object Binder {
+
+  /**
+   * trait for the context binding
```
trane: how can you make this comment more useful? right now it is just a description of the words used in line 14.

core/src/main/scala/com/lookout/borderpatrol/Binder.scala
```diff
@@ -0,0 +1,67 @@
...
+   * @tparam A
+   */
+  trait BinderContext[A] {
+    def name(manager: A): String
+    def hosts(manager: A): String
+  }
+
+  /**
+   * Bind request pod
```
trane: not sure that this means or how it describes what a `BindRequest` is.

core/src/main/scala/com/lookout/borderpatrol/Binder.scala
```diff
@@ -0,0 +1,67 @@
...
+  }
+  implicit object ServiceIdentifierBinderContext extends BinderContext[ServiceIdentifier] {
+    def name(sid: ServiceIdentifier): String = sid.name
+    def hosts(sid: ServiceIdentifier): String = sid.hosts
+  }
+
+  /**
+   * Binder objects
+   */
+  case class LoginManagerBinder() extends MBinder[LoginManager]
```
trane: why not make all of these `case object`s?

core/src/main/scala/com/lookout/borderpatrol/Binder.scala
```diff
@@ -0,0 +1,67 @@
...
+   */
+  trait BinderContext[A] {
+    def name(manager: A): String
+    def hosts(manager: A): String
+  }
+
+  /**
+   * Bind request pod
+   */
+  case class BindRequest[A](context: A, req: Request)
```
trane: alternatively you could do the following to clean up the `MBinder` class.

```
case class BindRequest[A : BinderContext](context: A, req: Request) {
  def name: String = implicitly[BinderContext[A]].name(context)
  def hosts: String = implicitly[BinderContext[A]].hosts(context)
}
```

core/src/main/scala/com/lookout/borderpatrol/ServiceMatcher.scala
```diff
@@ -43,10 +43,10 @@ case class ServiceMatcher(services: Set[ServiceIdentifier]) {
    * @param cmp A comparable function
    * @return The maximuma of folding over the set with the cmp function
    */
-  private[this] def foldWith(sis: Set[ServiceIdentifier],
-                             cmp: (ServiceIdentifier, ServiceIdentifier) => ServiceIdentifier):
-      Option[ServiceIdentifier] =
-    sis.foldRight(Option.empty[ServiceIdentifier])((lhs, res) => res match {
+  private[this] def foldWith[A](sis: Set[A],
+                             cmp: (A, A) => A):
```
trane: whitespace helping needed... you can now move this to one line.

maheshkelkar: done

core/src/main/scala/com/lookout/borderpatrol/Binder.scala
maheshkelkar: done

core/src/main/scala/com/lookout/borderpatrol/Binder.scala
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * It enables dynamic binding to the endpoints (e.g. login service, identity service, etc)
+   *
+   * Note that BinderContext makes it possible to templatize this code for all the LoginManagerBinder, ManagerBinder,
+   * ServiceIdentigfierBinder, etc, by making calls to methods (e.g. name & hosts) visible in the template.
+   *
+   * @param cache Caches the already established client service
+   * @tparam A
+   */
+  abstract class MBinder[A: BinderContext](cache: mutable.Map[String, Service[Request, Response]] =
```
trane: nit pick on whitespace:
 - `extends` should be an extra 2 spaces in from where `def` is if on a new line

maheshkelkar: done

core/src/main/scala/com/lookout/borderpatrol/Binder.scala
```diff
@@ -0,0 +1,71 @@
...
+                                       mutable.Map.empty[String, Service[Request, Response]])
+    extends Service[BindRequest[A], Response] {
+    def apply(req: BindRequest[A]): Future[Response] = {
+      cache.getOrElse(req.name,
```
trane: since `cache` is a parameter to the instantiation of this class, why not instantiate all of the services before-hand so that the cache become `immutable`?

core/src/main/scala/com/lookout/borderpatrol/Binder.scala
```diff
@@ -0,0 +1,71 @@
...
+        util.Combinators.tap(Httpx.newService(req.hosts))(cli => cache(req.name) = cli)
```
trane: if not, then you should synchronize and update the cache here

core/src/main/scala/com/lookout/borderpatrol/Binder.scala
```diff
@@ -0,0 +1,71 @@
...
+      ).apply(req.req)
```
trane: there is also: http://docs.scala-lang.org/overviews/collections/concrete-mutable-collection-classes.html#concurrent_maps

core/src/main/scala/com/lookout/borderpatrol/Binder.scala
maheshkelkar: @trane to do this, we will have to walk through entire config and populate cache with all the loginManager configs. 
advantage: we can change it to mutable
disadvantage: if we add config via REST API, this code will have to changed as well

maheshkelkar: Moved the code block to `this.synchronized`

maheshkelkar:
@PeterFCaswell Please take a look at this code review.

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -23,22 +24,25 @@ object Keymaster {
...
    * identity (master token)
-   * @param client Keymaster service
-   * @param path Keymaster service endpoint path
+   * @param binder Binder to bind to Keymaster identity service
    */
-  case class KeymasterIdentityProvider(client: Service[Request, Response], path: Path)
+  case class KeymasterIdentityProvider(binder: MBinder[Manager])
       extends IdentityProvider[Credential, Tokens] {
 
     def api(cred: Credential): Request =
```
PeterFCaswell: Should there be a comment on this method describing it and it's parameter?

bpConfig.json
```diff
@@ -0,0 +1,48 @@
+{
```
PeterFCaswell: Is there some sort of schema for the JSON here, or a document that talks about what these bits mean and what you can and can't have in it?

core/src/main/scala/com/lookout/borderpatrol/Binder.scala
```diff
@@ -0,0 +1,71 @@
+package com.lookout.borderpatrol
+
+import com.twitter.finagle.{Httpx, Service}
+import com.twitter.finagle.httpx.{Response, Request}
+import com.twitter.util.Future
+import scala.collection.mutable
+
+object Binder {
```
PeterFCaswell: What is a Binder? What does it do? Who might use it?
Would be good to see this in a comment.

core/src/main/scala/com/lookout/borderpatrol/crypto/Signer.scala
```diff
@@ -15,6 +15,9 @@ trait Signer {
   lazy val hmac: Mac = tap(Mac.getInstance(algo))(mac => mac.init(key))
 
   def sign(bytes: IndexedSeq[Byte]): Signature =
```
PeterFCaswell: Method comment?

bpConfig.json
maheshkelkar: No Schema. Unless there is a way to write that. I am going to update the README.md with a detailed information

core/src/main/scala/com/lookout/borderpatrol/Binder.scala
maheshkelkar: done

core/src/main/scala/com/lookout/borderpatrol/crypto/Signer.scala
maheshkelkar: I have fixed this with launching the hmac instance on the fly. Will add a comment

maheshkelkar:
@trane @PeterFCaswell I will push this as soon as all the checks pass. We can open a new issue, in case some of your concerns are not addressed

PR #86, opened 2015-11-05
Added BorderFilter to detect if session is tagged i.e. authenticated or not

### LINKED ISSUE #85: Detect authenticated sessions/users, trying to login again and redirect them to default page
- Detect if Session (i.e. user) is authenticated or not from the SessionId (i.e. add a new byte inside Sessionid)
- If SessionId is authenticated, then any HTTP requests to LoginManager paths (i.e. unauthenticated paths) should be redirected to the protected Service endpoint (i.e. `path` of the ServiceIdentifier)
- If SessionId is not authenticated, then any HTTP requests to protected endpoint paths, should be redirected to LoginManager path

maheshkelkar:
- Detect if Session (i.e. user) is authenticated or not from the SessionId
  (i.e. added a new byte inside Sessionid)
- If SessionId is authenticated, then any HTTP requests to LoginManager paths
  (i.e. unauthenticated paths) should be redirected to the protected
  Service endpoint (i.e. path of the ServiceIdentifier)
- If SessionId is not authenticated (i.e. not tagged), then any HTTP requests
  to protected endpoint paths, should be redirected to LoginManager path

Fixes #85

### DIFF
--- a/core/src/main/scala/com/lookout/borderpatrol/ServiceIdentifier.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/ServiceIdentifier.scala
@@ -16,8 +16,10 @@ import com.twitter.finagle.httpx.path.Path
  */
 case class ServiceIdentifier(name: String, hosts: Set[URL], path: Path, subdomain: String,
                              loginManager: LoginManager) {
-  def isMatchingPath(p: Path): Boolean = isServicePath(p) || isLoginManagerPath(p)
-  def isServicePath(p: Path): Boolean = p.startsWith(path)
+  def isMatchingPath(p: Path): Boolean =
+    isServicePath(p) || isLoginManagerPath(p)
+  def isServicePath(p: Path): Boolean =
+    p.startsWith(path)
   def isLoginManagerPath(p: Path): Boolean =
     !Set(loginManager.path, loginManager.loginPath).filter(p.startsWith(_)).isEmpty
 }
--- a/core/src/main/scala/com/lookout/borderpatrol/ServiceIdentifier.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/ServiceIdentifier.scala
@@ -12,7 +12,7 @@ import com.twitter.finagle.httpx.path.Path
  * @param hosts The list of URLs to upstream service
  * @param path The external url path prefix that routes to the internal service
  * @param subdomain A default fall-back when path is only `/`
- * @param loginManager The location to send a user when a request to this service is Unauthorized
+ * @param loginManager The location to send a user when a request to this service is Unauthenticated
  */
 case class ServiceIdentifier(name: String, hosts: Set[URL], path: Path, subdomain: String,
                              loginManager: LoginManager) {
--- a/core/src/main/scala/com/lookout/borderpatrol/ServiceIdentifier.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/ServiceIdentifier.scala
@@ -15,4 +15,9 @@ import com.twitter.finagle.httpx.path.Path
  * @param loginManager The location to send a user when a request to this service is Unauthorized
  */
 case class ServiceIdentifier(name: String, hosts: Set[URL], path: Path, subdomain: String,
-                             loginManager: LoginManager)
+                             loginManager: LoginManager) {
+  def isMatchingPath(p: Path): Boolean = isServicePath(p) || isLoginManagerPath(p)
+  def isServicePath(p: Path): Boolean = p.startsWith(path)
+  def isLoginManagerPath(p: Path): Boolean =
+    !Set(loginManager.path, loginManager.loginPath).filter(p.startsWith(_)).isEmpty
+}
--- a/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
@@ -69,15 +69,24 @@ case class SessionIdFilter(store: SessionStore)(implicit secretStore: SecretStor
  */
 class BorderFilter extends Filter[SessionIdRequest, Response, SessionIdRequest, Response] {
 
+  def servicePath(req: SessionIdRequest): Boolean =
+    req.req.serviceId.isServicePath(Path(req.req.req.path))
+
+  def redirectTo(path: Path): Response =
+    tap(Response(Status.Found))(res => res.location = path.toString)
+
+  def redirectToService(req: SessionIdRequest): Future[Response] =
+    redirectTo(req.req.serviceId.path).toFuture
+
+  def redirectToLogin(req: SessionIdRequest): Future[Response] =
+    redirectTo(req.req.serviceId.loginManager.path).toFuture
+
   def apply(req: SessionIdRequest, service: Service[SessionIdRequest, Response]): Future[Response] =
-    if (Tag.authenticated(req.sessionId.tag))
-      if (!req.req.serviceId.isServicePath(Path(req.req.req.path)))
-        tap(Response(Status.Found))(res => res.location = req.req.serviceId.path.toString).toFuture
-      else service(req)
-    else
-      if (req.req.serviceId.isServicePath(Path(req.req.req.path)))
-        tap(Response(Status.Found))(res => res.location = req.req.serviceId.loginManager.path.toString).toFuture
-      else service(req)
+    req.sessionId.tag match {
+      case AuthenticatedTag if !servicePath(req) => redirectToService(req)
+      case Untagged if servicePath(req) => redirectToLogin(req)
+      case _ => service(req)
+    }
 }
 
 /**
--- a/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
@@ -64,13 +64,13 @@ case class SessionIdFilter(store: SessionStore)(implicit secretStore: SecretStor
 /**
  * Determine if the session is tagged (i.e. authenticated) or not and force changes to request path
  * E.g.
- * - If SessionId is tagged (i.e. al, and the path is NOT a service path, then redirect it to service identifier path
- * - If SessionId is NOT tagged and path is a service path, then redirect to login page
+ * - If SessionId is authenticated and the path is NOT a service path, then redirect it to service identifier path
+ * - If SessionId is NOT authenticated and path is a service path, then redirect to login page
  */
 class BorderFilter extends Filter[SessionIdRequest, Response, SessionIdRequest, Response] {
 
   def apply(req: SessionIdRequest, service: Service[SessionIdRequest, Response]): Future[Response] =
-    if (SessionId.isTagged(req.sid))
+    if (SessionId.isAuthenticated(req.sid))
       if (!req.req.serviceId.isServicePath(Path(req.req.req.path)))
         tap(Response(Status.Found))(res => res.location = req.req.serviceId.path.toString).toFuture
       else service(req)
--- a/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
@@ -3,6 +3,7 @@ package com.lookout.borderpatrol.auth
 import com.lookout.borderpatrol.util.Combinators.tap
 import com.lookout.borderpatrol.{ServiceIdentifier, ServiceMatcher}
 import com.lookout.borderpatrol.sessionx._
+import com.twitter.finagle.httpx.path.Path
 import com.twitter.finagle.httpx.{Method, Status, Request, Response}
 import com.twitter.finagle.{Service, Filter}
 import com.twitter.util.Future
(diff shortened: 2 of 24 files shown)


maheshkelkar:
@PeterFCaswell Please review this change

core/src/main/scala/com/lookout/borderpatrol/ServiceIdentifier.scala
```diff
@@ -15,4 +15,9 @@ import com.twitter.finagle.httpx.path.Path
  * @param loginManager The location to send a user when a request to this service is Unauthorized
```
PeterFCaswell: Unauthorized or Unauthenticated?? I think you mean the latter.

PeterFCaswell: Or maybe you don't. :-( Maybe you might want to add a comment about this situation.

core/src/main/scala/com/lookout/borderpatrol/sessionx/Session.scala
```diff
@@ -80,8 +80,8 @@ object Session {
    * @tparam A
    * @return
    */
-  def apply[A](data: A)(implicit store: SecretStoreApi): Future[Session[A]] =
-    SessionId.next map (Session(_, data))
-
+  def apply[A](data: A, tagged: Boolean)(implicit store: SecretStoreApi): Future[Session[A]] =
```
PeterFCaswell: I'm wondering if 'tagged' should be a method on the request, rather than a separate parameter here and there?

core/src/main/scala/com/lookout/borderpatrol/sessionx/SessionId.scala
```diff
@@ -48,11 +50,11 @@ object SessionId {
   private[sessionx] def timeBytes(t: Time): TimeBytes =
     Injection.long2BigEndian(t.inMilliseconds)
 
-  private[sessionx] def payload(t: Time, e: Entropy, i: SecretId): Payload =
-    timeBytes(t) ++ e :+ i
+  private[sessionx] def payload(t: Time, e: Entropy, i: SecretId, tagId: TagId): Payload =
+    timeBytes(t) ++ e :+ i :+ tagId
```
PeterFCaswell: ++ and :+ aren't standard Scala operators, are they? If they're not, then this should have a comment about what is going on? (And perhaps where the operators come from?)

core/src/main/scala/com/lookout/borderpatrol/sessionx/SessionId.scala
```diff
@@ -108,14 +118,24 @@ object SessionId {
   def fromResponse(rep: Response)(implicit ev: SecretStoreApi): Try[SessionId] =
     fromCookie(rep.cookies.get("border_session"))
 
+  /**
+   * Tagging can be used to distiguish a set of sessionId.
+   * For e.g. tagged Sessionid could be used to represent authenticated, host specific, etc. SessionIds
```
PeterFCaswell: Is there a list somewhere of the current tags?

core/src/main/scala/com/lookout/borderpatrol/sessionx/SessionId.scala
```diff
@@ -32,6 +32,8 @@ object SessionId {
 
   val entropySize: Size = 16
   val lifetime = Duration(1, TimeUnit.DAYS)
+  val nullTagId: TagId = 0.toByte
```
PeterFCaswell: Probably want a comment here indicating that this is forming a list of possible TagIds and noting whether this is a bit-field or an integer.

PeterFCaswell:
Mostly just comment requests. Feel free to merge after you've added what makes sense.

core/src/main/scala/com/lookout/borderpatrol/ServiceIdentifier.scala
maheshkelkar: unauthenticated

core/src/main/scala/com/lookout/borderpatrol/sessionx/Session.scala
maheshkelkar: ok, Let me experiment with applyTagged() method

core/src/main/scala/com/lookout/borderpatrol/sessionx/SessionId.scala
maheshkelkar: Both of these are scala collections operators. But, `timeBytes(t) and `e` are `IndexedSeq[Bytes]`, whereas `i` & `tagId` are bytes. In general its just appending going on. I will add a comment

core/src/main/scala/com/lookout/borderpatrol/sessionx/SessionId.scala
maheshkelkar: done

core/src/main/scala/com/lookout/borderpatrol/sessionx/SessionId.scala
```diff
@@ -33,6 +34,13 @@ object SessionId {
   val entropySize: Size = 16
   val lifetime = Duration(1, TimeUnit.DAYS)
 
+  /**
+   * TagId is a byte that differentiates the set of SessionId from others.
+   * Currently we are using a single bit for differentiation. Declare the constants here.
+   */
+  val nullTagId: TagId = 0.toByte /* Not tagged */
```
trane: why not make this an algebraic data type?

```
trait TagId {
  val id: Byte
}

case object NullTag extends TagId {
  val id = 0.toByte
}

case object AuthenticatedTag extends TagId {
  val id = 1.toByte
}
```

then you use this later like:
```
def someFooMethod(tagId: TagId) {
  SessionId(???, ???, ???, tagId.id)
}
```

and your checks become compile-time checks!

you could remove the `isTagged` and `isAuthenticated` methods now, since we have types to identify the tags:

```
tagId match {
  case NullTag => ???
  case AuthenticatedTag => ???
}
```

core/src/main/scala/com/lookout/borderpatrol/sessionx/SessionId.scala
```diff
@@ -69,14 +77,14 @@ object SessionId {
    * @param store where to fetch the current [[Secret]] to sign this id
    * @return
    */
-  def next(implicit store: SecretStoreApi): Future[SessionId] =
-    (SessionId(currentExpiry, genEntropy, store.current)).toFuture
+  def next(tagId: TagId = nullTagId)(implicit store: SecretStoreApi): Future[SessionId] =
```
trane: i'm not a huge fan of requiring a `()`

we could have an api like `SessionId.authenticated` and `SessionId.null`, which we could implement by adding the methods to this object.

If you use types, then we could make `next` have a parameterized type, `next[AuthenticatedTag]` or `next[NullTag]`

we could allow for chaining of methods... `SessionId.next.withTag(AuthenticatedTag)` and `SessionId.next.withTag(NullTag)` for null. this would be implemented like this (assuming you are using tag types):

```
def next(implicit store: SecretStoreApi): TagId => Future[SessionId] = t =>
  (SessionId(currentExpiry, genEntropy, store.current, t.id)).toFuture

def withTag(tagId: TagId)(f: TagId => Future[SessionId]): Future[SessionId] =
  f(tagId)
```

core/src/main/scala/com/lookout/borderpatrol/sessionx/SessionId.scala
```diff
@@ -33,6 +34,13 @@ object SessionId {
   val entropySize: Size = 16
   val lifetime = Duration(1, TimeUnit.DAYS)
 
+  /**
+   * TagId is a byte that differentiates the set of SessionId from others.
+   * Currently we are using a single bit for differentiation. Declare the constants here.
+   */
+  val nullTagId: TagId = 0.toByte /* Not tagged */
+  val authenticatedTagId: TagId = 1.toByte /* is session authenticated */
```
trane: if we are calling this a `Tag` then we should probably call `NullTag` something else or have a `LoginTag` or something.

core/src/main/scala/com/lookout/borderpatrol/ServiceIdentifier.scala
```diff
@@ -12,7 +12,12 @@ import com.twitter.finagle.httpx.path.Path
  * @param hosts The list of URLs to upstream service
  * @param path The external url path prefix that routes to the internal service
  * @param subdomain A default fall-back when path is only `/`
- * @param loginManager The location to send a user when a request to this service is Unauthorized
+ * @param loginManager The location to send a user when a request to this service is Unauthenticated
  */
 case class ServiceIdentifier(name: String, hosts: Set[URL], path: Path, subdomain: String,
-                             loginManager: LoginManager)
+                             loginManager: LoginManager) {
+  def isMatchingPath(p: Path): Boolean = isServicePath(p) || isLoginManagerPath(p)
```
trane: i've tried to use the following convention:

```
val stuff: String = "on the same line"
val valFunc: String => String = s => "on the same line, too even if it's a function"

def func(thingy: Thing): String =
  "body is on the next line, even if it's short"
```

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -61,6 +62,25 @@ case class SessionIdFilter(store: SessionStore)(implicit secretStore: SecretStor
...
 /**
+ * Determine if the session is tagged (i.e. authenticated) or not and force changes to request path
+ * E.g.
+ * - If SessionId is authenticated and the path is NOT a service path, then redirect it to service identifier path
+ * - If SessionId is NOT authenticated and path is a service path, then redirect to login page
+ */
+class BorderFilter extends Filter[SessionIdRequest, Response, SessionIdRequest, Response] {
+
+  def apply(req: SessionIdRequest, service: Service[SessionIdRequest, Response]): Future[Response] =
+    if (Tag.authenticated(req.sessionId.tag))
```
trane: why not...

```
// add some methods
def servicePath(req: Request): Boolean =
  req.req.serviceId.isServicePath(Path(req.req.req.path))

def redirectTo(path: Path): Unit =
  tap(Response(Status.Found))(res => res.location = path.toString)

def redirectToService(req: Request): Future[Response] =
  redirectTo(req.req.serviceId.path).toFuture

def redirectToLogin(req: Request): Future[Response] =
  redirectTo(req.req.serviceId.loginManager.path).toFuture

// then apply is much nicer like this...
def apply(req: SessionIdRequest, service: Service[SessionIdRequest, Response]): Future[Response] = req.sessionId.tag match {
  case AuthenticatedTag if !servicePath(req) => redirectToService(req)
  case Untagged if servicePath(req) => redirectToLogin(req)
  case _ => service(req)
}
```

core/src/main/scala/com/lookout/borderpatrol/crypto/Signer.scala
```diff
@@ -19,6 +20,6 @@ trait Signer {
    * @return Signature
    */
   def sign(bytes: IndexedSeq[Byte]): Signature =
-    tap(Mac.getInstance(algo))(mac => mac.init(key)).doFinal(bytes.toArray).toIndexedSeq
+   synchronized(hmac.doFinal(bytes.toArray).toIndexedSeq)
```
trane: i was thinking about this some more, and i've convinced myself i'm wrong about using synchronized. 

- the time it takes to initialize the Mac should be dwarfed by the time it takes to sign the payload.
- the synchronize means that we are limiting almost all incoming requests to a single thread (since every request is verified)

this means, by using synchronize, we are gaining much less than we are losing.

i recommend changing the `hmac` from a `lazy val` to a `def`, and remove the `synchronized`

```
def hmac: Mac =
  tap(Mac.getInstance(algo))(mac => mac.init(key))
```

core/src/main/scala/com/lookout/borderpatrol/sessionx/Session.scala
```diff
@@ -78,10 +78,10 @@ object Session {
    * @param data value you want to store
    * @param store the secret store to fetch current secret
    * @tparam A
-   * @return
+   * @return Session
    */
-  def apply[A](data: A)(implicit store: SecretStoreApi): Future[Session[A]] =
-    SessionId.next map (Session(_, data))
-
+  def apply[A](data: A, tag: Tag = Untagged)(implicit store: SecretStoreApi): Future[Session[A]] =
```
trane: this is an api that would change if we allowed for different `Session` types (from the gist)

```
// "initial session" == Session[Request] := InitialSessionId x Data[Request]
// "authenticated session" == Session[?] := AuthenticatedSessionId x Data[?]
```

core/src/main/scala/com/lookout/borderpatrol/sessionx/Tag.scala
```diff
@@ -0,0 +1,29 @@
+package com.lookout.borderpatrol.sessionx
```
trane: where is the TagSpec?

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
maheshkelkar: very slick

core/src/main/scala/com/lookout/borderpatrol/ServiceIdentifier.scala
maheshkelkar: done

core/src/main/scala/com/lookout/borderpatrol/crypto/Signer.scala
maheshkelkar: done

core/src/main/scala/com/lookout/borderpatrol/sessionx/Tag.scala
maheshkelkar: done

maheshkelkar:
@trane I have updated the code, let me know if it looks ok. If yes, I will merge it

PR #87, opened 2015-11-09
Create a consul backed secret store api class and tests

tejom:
-Add a SecretsCache to handle secrets from Consul

 -Add class ConsulConnection to connect and make requests to a Consul kv store

 -Create json encoder for Secrets and consul kv store replies

 -Add tests for new api class

For issue #40

### DIFF
--- a/core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretsCache.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretsCache.scala
@@ -1,0 +1,165 @@
+package com.lookout.borderpatrol.sessionx
+
+import argonaut.Argonaut._
+import argonaut.{CodecJson, Parse}
+import com.twitter.finagle.{httpx, Service}
+import com.twitter.io.Buf
+import com.twitter.util.Future
+import scala.util.{Try, Failure, Success}
+import scalaz._
+
+
+/**
+ * Stores the secrets stored on the consul server.
+ * Polls the consul server and keeps an inmemory cache of Secrets
+ *
+ * @param poll How often in seconds to check for updates
+ * @param consul An instance on ConsulConnection to make connections with the consul server
+ */
+
+case class ConsulSecretCache(poll: Long, consul: ConsulConnection, secretsStart: Secrets) extends Runnable {
+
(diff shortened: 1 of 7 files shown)


core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretsCache.scala
```diff
@@ -0,0 +1,165 @@
...
+  }
+}
+
+/**
+ * class to interface with consul kv store
+ *
+ * @param consul Finagle server to send and get requests to the server
+ * @param host host name of the consul server to connect to ex: "localhost"
+ */
+case class ConsulConnection(consul: Service[httpx.Request, httpx.Response], host: String, port: String) {
```
maheshkelkar: why do we need ConsulConnection, why can't this be a simple httpx client? 
1. All the json encoding/decoding should be in one place and 
2. we don't need `value` function, because we can just move this code to a place where we are consuming the response

maheshkelkar:
@tejom Thank you for doing this work. My main concern is there are too many classes/objects, ConsulSecretStore, ConsulSecretCache, ConsulConnection, ConsulResponse, etc. Do we really need all of them? Is there a way to collapse them into oneanother? That will make the code compact.

However, if this code is working, lets commit it for the timebeing.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretsCache.scala
tejom: I made a ConsulConnnection class to make handling the consul specific requests a little easier. Consul's kv store does a bunch of things to a value thats added so I thought putting all of this into a class makes it more organized. /v1/kv is also part of consul's kv store so instead of writing that as part of every request I thought putting it into one place would make it easier. 

I thought the function `value` would imply just getting the value from the key passed. Otherwise getting a response from consul just using http request would get a list of json that includes a lot of consul specific information along with the value. That value is also encoded in base64 and needs to be decoded into something readable. 

@trane was talking to me about rewriting `set` to not need the url and port. So if thats done then maybe the class can be turned into a object with some helper functions or something?

trane:
@tejom can you please resolve the merge conflicts so that I can merge this?

PR #88, opened 2015-11-10
Resurrected server module with code that weaves core and auth module together

### LINKED ISSUE #52: borderpatrol-server module

maheshkelkar:
- Moved ServerConfig and Services code to Server module
- cleaned up old unused files, code

Fixes #52


server/src/main/scala/com/lookout/borderpatrol/server/services.scala
```diff
@@ -1,14 +1,72 @@
...
-  case class AuthService()
+  /**
+   * Glue that DEMUXes the chain into appropriate identityProvider or accessIssuer
+   *
+   * @param accessIssuerMap
+   * @param identityProviderMap
+   */
+  case class MainGlueService(identityProviderMap: Map[String, Service[SessionIdRequest, Response]],
+                             accessIssuerMap: Map[String, Service[SessionIdRequest, Response]])
+    extends Service[SessionIdRequest, Response] {
```
trane: indent 4 spaces from start of `case class`

server/src/main/scala/com/lookout/borderpatrol/server/services.scala
```diff
@@ -1,14 +1,72 @@
...
 object services {
 
-  case class AuthService()
+  /**
+   * Glue that DEMUXes the chain into appropriate identityProvider or accessIssuer
+   *
+   * @param accessIssuerMap
+   * @param identityProviderMap
+   */
+  case class MainGlueService(identityProviderMap: Map[String, Service[SessionIdRequest, Response]],
```
trane: this is low-hanging fruit ripe for refactoring, but it works for now.

server/src/main/scala/com/lookout/borderpatrol/server/services.scala
```diff
@@ -1,14 +1,72 @@
...
+  /**
+   * The sole entry point for all service chains
+   */
+  def MainServiceChain(implicit config: ServerConfig): Service[Request, Response] = {
+    implicit val secretStore = config.secretStore
+    val serviceMatcher = ServiceMatcher(config.serviceIdentifiers)
+    val identityProviderMap = Map("keymaster" -> keymasterIdentityProviderChain(config.sessionStore))
+    val accessIssuerMap = Map("keymaster" -> keymasterAccessIssuerChain(config.sessionStore))
 
+    new ExceptionFilter andThen /* Convert exceptions to responses */
```
trane: any reason you are using `new` over the `apply` syntax? it's not as scala-esque to use `new` when you don't need to

what's stopping you from doing this?...
```
ExceptionFilter() andThen 
ServiceFilter(serviceMatcher) andThen 
SessionIdFilter(config.sessionStore) andThen 
BorderFilter() andThen 
MainGlueService(identifyProviderMap, accessIssuerMap)
```

server/src/main/scala/com/lookout/borderpatrol/server/services.scala
```diff
@@ -1,14 +1,72 @@
...
+      accessIssuerMap.get(accessManager.name) match {
+        case Some(s) => s
+        case None => throw InvalidConfigError("Failed to find AccessIssuer Service Chain for " +
+          accessManager.name)
+      }
+
+    def apply(req: SessionIdRequest): Future[Response] = {
+      val p = Path(req.req.req.path)
+      /* Upstream service path */
+      if (p.startsWith(req.req.serviceId.path))
```
trane: isn't this just the `BorderFilter` implemented slightly differently?

interesting api change 
  - if we define `SessionIdRequest` as `Authenticated` vs `Initial` 
  - construction of the `SessionIdRequest` would depend on the path, like `BorderFilter`

then we could make all of these `if/else if/else` statements part of a type-based api of a `SessionIdRequest`.

anyway... some psuedo code with a basic shape:

```
sealed trait SessionIdRequest {
  val id: SessionId
  val serviceId: ServiceIdentifier
  val req: Request
}
case class AuthenticatedRequest(id: SessionId, req: Request, serviceId: ServiceIdentifier) extends SessionIdRequest
case class InitialRequest(id: SessionId, req: Request, serviceId: ServiceIdentifier) extends SessionIdRequest

object SessionIdRequest {
  def apply(req: Request, id: SessionId, serviceId: ServiceIdentifer): SessionIdRequest =
    id.tag match {
      case Authenticated => AuthenticatedRequest(req, id, serviceId)
      case Initial => InitialRequest(req, id, serviceId)
    }
}

case class MainGlueService... {
  def accessIssuer(req: AuthenticatedRequest): Service[AuthenticatedRequest, Response] =
    accessIssuerMap.get(req.serviceId.accessManager.name) getOrElse throw InvalidConfigError("...")
  
  def identityProvider(req: InitialRequest): Service[InitialRequest, Response] =
    identityProviderMap.get(req.serviceId.identityManager.name)(req) getOrElse throw InvalidConfigError("...")

  def apply(req: SessionIdRequest): Future[Response] = req match {
     case AuthenticatedRequest => accessIssuer(req)(req)
     case InitialRequest => identityProvider(req)(req)
     case _ => Response(Status.NotFound).toFuture
  }
}

File: server/src/main/scala/com/lookout/borderpatrol/server/services.scala [hunk 5]
```diff
@@ -1,14 +1,72 @@
...
+   * @param identityProviderMap
+   */
+  case class MainGlueService(identityProviderMap: Map[String, Service[SessionIdRequest, Response]],
+                             accessIssuerMap: Map[String, Service[SessionIdRequest, Response]])
+    extends Service[SessionIdRequest, Response] {
+
+    def getIdentityProviderService(identityManager: Manager): Service[SessionIdRequest, Response] =
+      identityProviderMap.get(identityManager.name) match {
+        case Some(s) => s
+        case None => throw InvalidConfigError("Failed to find IdentityProvider Service Chain for " +
```
trane: shouldn't this return a `Service[SessionIdRequest, Response]` that always returns a failure `Response` in json?

```
case class UnidentifiedService(name: String) extends Service[SessionIdRequest, Response] {
  def apply(req: Request): Future[Response] =
    tap(Response(Status.NotFound))(res => res.contentString = """{"error": "$name"}""").toFuture
}
```

then this becomes...
```
def identityProvider(manager: Manager): Service[SessionIdRequest, Response] = 
  identityProviderMap.get(manager.name) getOrElse UnidentifiedService(s"Failed to find IdentityProvider for ${manager.name}")
```

server/src/main/scala/com/lookout/borderpatrol/server/services.scala
```diff
@@ -1,14 +1,72 @@
...
+   * Glue that DEMUXes the chain into appropriate identityProvider or accessIssuer
+   *
+   * @param accessIssuerMap
+   * @param identityProviderMap
+   */
+  case class MainGlueService(identityProviderMap: Map[String, Service[SessionIdRequest, Response]],
+                             accessIssuerMap: Map[String, Service[SessionIdRequest, Response]])
+    extends Service[SessionIdRequest, Response] {
+
+    def getIdentityProviderService(identityManager: Manager): Service[SessionIdRequest, Response] =
```
trane: why `getXXXX`? we don't really use `get` in Scala much, unless you are providing a Java-esque api
why `xxxxService`? doesn't the return type showing `Service` give us enough info?

why not just `identity` and `access`? or `identityProvider` and `accessIssuer`?

server/src/main/scala/com/lookout/borderpatrol/server/services.scala
maheshkelkar: Let me look into merging BorderFilter and MainGlueService. Because essentially its doing the same thing

maheshkelkar:
@trane incorporated all the comments, please take a look.

PR #91, opened 2015-11-12
Add assembly settings for building fat jar

maheshkelkar:


maheshkelkar:
@trane assuming that you are ok this, merging it now :)

PR #92, opened 2015-11-20
Add config to instantiate ConsulSecretStore

### LINKED ISSUE #40: Consul backed Secret Store

maheshkelkar:
- Also updated the README.md for secret and session store config
  examples
- Add version to fat jar

Fixes #40

### DIFF
--- a/server/src/test/scala/com/lookout/borderpatrol/server/ConfigSpec.scala
+++ b/server/src/test/scala/com/lookout/borderpatrol/server/ConfigSpec.scala
@@ -71,13 +71,12 @@ class ConfigSpec extends BorderPatrolSuite {
   it should "uphold encoding/decoding ServerConfig" in {
     def encodeDecode(config: ServerConfig) : ServerConfig = {
       val encoded = config.asJson
-      println("***MVK: encoded = " + encoded)
       decode[ServerConfig](encoded.toString()) match {
         case Xor.Right(a) => a
         case Xor.Left(b) => ServerConfig(defaultSecretStore, defaultSessionStore, Set(), Set(), Set(), Set())
       }
     }
-    //verifyServerConfig(encodeDecode(serverConfig), serverConfig)
+    verifyServerConfig(encodeDecode(serverConfig), serverConfig)
     verifyServerConfig(encodeDecode(serverConfig1), serverConfig1)
   }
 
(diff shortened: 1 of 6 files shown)


server/src/test/scala/com/lookout/borderpatrol/server/ConfigSpec.scala
```diff
@@ -70,13 +71,13 @@ class ConfigSpec extends BorderPatrolSuite {
   it should "uphold encoding/decoding ServerConfig" in {
     def encodeDecode(config: ServerConfig) : ServerConfig = {
       val encoded = config.asJson
+      println("***MVK: encoded = " + encoded)
```
trane: println?

server/src/test/scala/com/lookout/borderpatrol/server/ConfigSpec.scala
```diff
@@ -70,13 +71,13 @@ class ConfigSpec extends BorderPatrolSuite {
...
       decode[ServerConfig](encoded.toString()) match {
         case Xor.Right(a) => a
         case Xor.Left(b) => ServerConfig(defaultSecretStore, defaultSessionStore, Set(), Set(), Set(), Set())
       }
     }
-    val foo = encodeDecode(serverConfig)
-    verifyServerConfig(encodeDecode(serverConfig), serverConfig)
+    //verifyServerConfig(encodeDecode(serverConfig), serverConfig)
```
trane: why comment this out?

maheshkelkar: fixed. mistake

server/src/test/scala/com/lookout/borderpatrol/server/ConfigSpec.scala
maheshkelkar: debugging. took it out

PR #94, opened 2015-11-24
Add ProtoManager for protocol specific config

### LINKED ISSUE #93: Add protocol specific configuration to LoginManager
Move authentication protocol/framework specific configuration from `LoginManager` to `ProtoManager`

maheshkelkar:
- ProtoManager contains seperate config for each protocol
  (such as internal keymaster specific or OAuth2 or SAML)

Fixes #93

### DIFF
--- a/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
@@ -55,7 +55,8 @@ case class SessionIdFilter(store: SessionStore)(implicit secretStore: SecretStor
           session <- Session(req.req)
           _ <- store.update(session)
         } yield tap(Response(Status.Found)) { res =>
-          res.location = req.serviceId.loginManager.path.toString
+          res.location = req.serviceId.loginManager.protoManager.redirectLocation(
+            req.req.host.getOrElse("lookout.com"))
           res.addCookie(session.id.asCookie)
         }
     }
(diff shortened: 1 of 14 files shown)


core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -55,7 +55,8 @@ case class SessionIdFilter(store: SessionStore)(implicit secretStore: SecretStor
           session <- Session(req.req)
           _ <- store.update(session)
         } yield tap(Response(Status.Found)) { res =>
-          res.location = req.serviceId.loginManager.path.toString
+          res.location = req.serviceId.loginManager.protoManager.redirectLocation(
+            req.req.host.getOrElse("lookout.com"))
```
trane: shouldn't `"lookout.com"` come from config?

server/src/main/scala/com/lookout/borderpatrol/server/Config.scala
```diff
@@ -248,7 +295,8 @@ object Config {
   def readServerConfig(filename: String) : ServerConfig = {
     decode[ServerConfig](Source.fromFile(filename).mkString) match {
       case Xor.Right(a) => validate(a); a
-      case Xor.Left(b) => throw ConfigError("Failed to decode following fields: " +
+      case Xor.Left(b) =>
+        throw ConfigError("Failed to decode following fields: " +
```
trane: can't wait until we do this throwing at a different location :) already excited about 0.2.0 :)

trane:
other than the magic "lookout.com" strings, this looks ok to move forward.

PR #99, opened 2015-12-07
Add modern security headers

trane:


security/src/main/scala/com/lookout/borderpatrol/security/Headers.scala
```diff
@@ -0,0 +1,51 @@
...
+  val localIp = InetAddress.getLocalHost.getHostAddress
+
+  def apply(req: Request, service: Service[Request, Response]): Future[Response] =
+    service(tap(req) { r =>
+        r.xForwardedFor = (r.xForwardedFor match {
+          case Some(s) => s"$s, $localIp"
+          case None => localIp
+        })
+      r.headerMap ++= headers
+    })
```
maheshkelkar: I thought that HeaderFilter is supposed to add this headers to response going back to the client. But, perhaps I misunderstood.

But, are we adding this security headers to the request to the upstream services.

core/src/test/scala/com/lookout/borderpatrol/test/package.scala
```diff
@@ -20,4 +22,11 @@ package object test {
     def isThrowable: Boolean =
       Await.ready(future).poll.get.isThrow
   }
+
+  def testService(f: Request => Boolean): Service[Request, Response] = Service.mk[Request, Response](r => {
```
maheshkelkar: why not put this into HeadersSpec.scala, unless of course its used anywhere else

security/src/main/scala/com/lookout/borderpatrol/security/Headers.scala
```diff
@@ -0,0 +1,51 @@
...
+  *   X-Frame-Options: DENY
+  *   X-XSS-Protection: 1; mode=block
+  *   X-ContentType-Options: nosniff
+  *   X-Download-Options: noopen
+  *   X-Permitted-Cross-Domain-Policies: none
+  *   X-Forwarded-For: the ip of this host appended to any existing values
+  *
+  * @param headers
+  */
+case class HeaderFilter(headers: HeaderMap = Headers.all) extends SimpleFilter[Request, Response] {
```
maheshkelkar: Where are we using this? Shouldn't we integrate this in MainSeviceChain?

maheshkelkar:
@trane I am done with the review

core/src/test/scala/com/lookout/borderpatrol/test/package.scala
trane: Why would I put this into HeadersSpec? It's a completely general purpose service that allows us to test entire filter chains. We aren't using it anywhere else - *yet*!

security/src/main/scala/com/lookout/borderpatrol/security/Headers.scala
trane: yes, absolutely. this PR isn't to add this into the main chain, it's a module that can be loaded. it's up to you where it gets integrated :)

security/src/main/scala/com/lookout/borderpatrol/security/Headers.scala
trane: correct, these headers are for the request going to the services.
it might be nice to see which headers should/could be added to send back to the client in a response, though.

maheshkelkar: I will read about this, but not sure how do this header help us. 

Does the config or something else help us make a decision on what will add the headers to the response as well? I mean do we need investigation, discussion with Toan or a config knob?

PR #100, opened 2015-12-11
Add StatsdExporter for reporting metrics

### LINKED ISSUE #61: Add statsd listener for twitter Metrics
collect metrics from TwitterServer and Report it to datadog

maheshkelkar:
Fixes #61

### DIFF
--- a/server/src/main/scala/com/lookout/borderpatrol/server/StatsdExporter.scala
+++ b/server/src/main/scala/com/lookout/borderpatrol/server/StatsdExporter.scala
@@ -1,0 +1,93 @@
+package com.lookout.borderpatrol.server
+
+import java.net.InetAddress
+import java.nio.ByteBuffer
+import java.nio.channels.DatagramChannel
+import com.twitter.common.metrics.Metrics
+import com.twitter.finagle.stats._
+import com.twitter.finagle.util.{InetSocketAddressUtil, DefaultTimer}
+import com.twitter.io.Buf
+import com.twitter.util.{Duration, Timer, NonFatal}
+import scala.collection.JavaConverters.mapAsScalaMapConverter
+import scala.collection.Map
+
+
+class StatsdExporter(registry: Metrics, timer: Timer, prefix: String = "", duration: Duration, hostAndPort: String) {
+  private[this] val addr = InetSocketAddressUtil.parseHosts(hostAndPort).head
+  private[this] val channel = DatagramChannel.open()
+
+  // Alternate constructor
+  def this(config: StatsdExporterConfig) = this(MetricsStatsReceiver.defaultRegistry, DefaultTimer.twitter,
+    config.prefix.getOrDefault(InetAddress.getLocalHost.getHostName), Duration.fromSeconds(config.durationInSec),
+    config.host)
+
+  // Schedule exporter
+  timer.schedule(duration)(report)
+
+  // Format helpers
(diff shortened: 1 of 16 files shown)


server/src/main/scala/com/lookout/borderpatrol/server/StatsdExporter.scala
```diff
@@ -0,0 +1,93 @@
...
+
+  // Send helpers
+  private[this] def buf(utf8: String): Buf =
+    Buf.Utf8(utf8)
+
+  private[this] def byteBuf(buf: Buf): ByteBuffer =
+    Buf.ByteBuffer.Owned.extract(buf)
+
+  private[this] def send(str: String): Unit = {
+    println(str)
```
trane: `println`?

maheshkelkar: debugging. will take it out.

server/src/main/scala/com/lookout/borderpatrol/server/StatsdExporter.scala
```diff
@@ -0,0 +1,93 @@
...
+    channel.send(byteBuf(buf(str)), addr)
+  }
+
+  // Computer helpers
+  private[this] def labelPercentile(p: Double): String = {
+    // this has a strange quirk that p999 gets formatted as p9990
+    val gname: String = "p" + (p * 10000).toInt
```
trane: isn't that because you are multiplying by `10000` instead of `1000`?

server/src/main/scala/com/lookout/borderpatrol/server/package.scala
```diff
@@ -2,4 +2,12 @@ package com.lookout.borderpatrol
 
 package object server {
 
+  implicit class StringOps(val s: String) extends AnyVal {
```
trane: look at you @maheshkelkar making infix operators! :) :) :) :+1:

server/src/main/scala/com/lookout/borderpatrol/server/StatsdExporter.scala
maheshkelkar: This computation allows 0.99 -> p99, 0.95 -> p95 and 0.999 -> p9990.
Yes multiplying by 10000 makes that happen

trane: why not just do this then? `p.toString.replace("0.", "p")`

maheshkelkar: we have to convert following {.50, .90, .95, .99, .999, .9999} to {p50, p90, p95, p99, p999, p9999}
Your solution works for everything except (p50, p90, which gets printed as p5 & p9

trane: you could do the following:

```
val s = p.toString.replace("0.", "p")
if (s.size == 2) s"$s0"
else s
```

alternatively you can just make it a match statement:

```
def format(d: Double): String = d match {
  case .5 => "p50"
  case .9 => "p90"
  case s => s.toString.replace("0.", "p")
}

or...

def format(d: Double): String = d.toString.replace("0.", "p") match {
  case "p5" => "p50"
  case "p9" => "p90"
  case p => p
}
```

PR #104, opened 2015-12-15
Add rewritePath to ServiceIdentifier

### LINKED ISSUE #103: Add rewrite url capability to protected SeviceIdentifier
If internal & external paths are different, then this will allow us to hide the internal path from outside

maheshkelkar:
Its an optional parameter

If present, it will replaces the external path configuration in
ServiceIdentifier with the rewritePath configuration in the ServiceIdentifier

Fixes #103

### DIFF
--- a/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
@@ -179,7 +179,12 @@ case class AccessFilter[A, B](binder: MBinder[ServiceIdentifier])
             accessService: Service[AccessRequest[A], AccessResponse[B]]): Future[Response] =
     accessService(AccessRequest(req.id, req.req.req.serviceId, req.req.sessionId)).flatMap(accessResp =>
       binder(BindRequest(req.req.req.serviceId,
-        tap(req.req.req.req) { r => r.headerMap.add("Auth-Token", accessResp.access.access.toString)}))
+        tap(req.req.req.req) { r => {
+          // Rewrite the URI (i.e. path)
+          r.uri = req.req.req.serviceId.rewritePath.fold(r.uri)(p =>
+            r.uri.replaceFirst(req.req.req.serviceId.path.toString, p.toString))
+          r.headerMap.add("Auth-Token", accessResp.access.access.toString)
+        }}))
     )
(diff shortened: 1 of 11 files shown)


core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -179,7 +179,12 @@ case class AccessFilter[A, B](binder: MBinder[ServiceIdentifier])
             accessService: Service[AccessRequest[A], AccessResponse[B]]): Future[Response] =
     accessService(AccessRequest(req.id, req.req.req.serviceId, req.req.sessionId)).flatMap(accessResp =>
       binder(BindRequest(req.req.req.serviceId,
-        tap(req.req.req.req) { r => r.headerMap.add("Auth-Token", accessResp.access.access.toString)}))
+        tap(req.req.req.req) { r => {
+          // Rewrite the URI (i.e. path)
+          r.uri = req.req.req.serviceId.rewritePath.fold(r.uri)(p =>
```
trane: let's create a `RewriteFilter` instead.

maheshkelkar: hrm I thought about it, but I was not sure if we need a filter for 1 line of code. But, doing so, will provide flexibility, if somebody doesnt want it

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -179,15 +182,33 @@ case class AccessFilter[A, B](binder: MBinder[ServiceIdentifier])
...
 
 /**
+ * This filter rewrites Request Path as per the ServiceIdentifier configuration
+ */
+case class RewriteFilter() extends SimpleFilter[SessionIdRequest, Response] {
+  def apply(req: SessionIdRequest,
+            service: Service[SessionIdRequest, Response]): Future[Response] = {
+    service(new SessionIdRequest(req.req.serviceId, req.sessionId, tap(req.req.req) { r =>
+      // Rewrite the URI (i.e. path)
+      r.uri = req.req.serviceId.rewritePath.fold(r.uri)(p =>
```
trane: :)

PR #107, opened 2015-12-16
Add more debug level logs

maheshkelkar:


core/src/main/scala/com/lookout/borderpatrol/sessionx/SessionId.scala
```diff
@@ -25,7 +25,9 @@ import scala.util.{Success, Failure, Try}
  * @param secret the secret used to sign this session id
  * @param signature the bytes of the signature(expires, entropy, secret.id)
  */
-case class SessionId(expires: Time, entropy: Entropy, secret: Secret, tag: Tag, signature: Signature)
+case class SessionId(expires: Time, entropy: Entropy, secret: Secret, tag: Tag, signature: Signature) {
+  override def toString: String = SessionId.toBase64(this)
```
trane: i don't know that i agree with this. why do we want to make the `toString` method return the `base64` encoded version of itself? if this is for better logging, why are we logging these in full? if we want it for logging, it would be better to only log part of the signature, as it can't be used to gain access to memcached values.

maheshkelkar: Yes, this is for better logging. ok, let me change this to signature.

PR #110, opened 2016-01-10
Download AAD certificate and verify signature on Access Token

### LINKED ISSUE #109: BorderPatrol should verify the signature on AAD AccessToken
- For OAuth2 protocol, BP should verify signature on the AccessToken. 
- To do so, it should download the certificate from the endpoint (configuration) and use public key from it. 
- Perhaps it should cache the certificate as well

maheshkelkar:
- Also, split KeymasterPreLoginFilter to KeymasterTransformFilter and
  KeymasterPreLoginFilter
- Moved oAuth2 functionality to core
- Added caching for AAD certificate (still missing timely gc)
- Support for RSA & EC algorithms

Fixes #109


auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -79,6 +80,41 @@ object Keymaster {
   }
 
   /**
+   * Handles Keymaster transforms for internal and OAuth2
+   */
+  case class KeymasterTransformFilter(oAuth2CodeVerify: OAuth2CodeVerify)(implicit statsReceiver: StatsReceiver)
```
trane: is the idea that - via construction - we will just keep adding `transform` methods for each type and case match on the `ProtoManager`? so, next version for authentication named `BazAuth` will be added to this constructor and a method called `transformBazAuth` will be added, too?

auth/src/test/scala/com/lookout/borderpatrol/auth/keymaster/KeymasterSpec.scala
```diff
@@ -18,11 +19,12 @@ import com.twitter.finagle.httpx.path.Path
 import com.twitter.finagle.httpx._
 import com.twitter.util.{Await, Future}
 
+import org.scalatest.mock.MockitoSugar
+import org.mockito.Mockito._
 
-class KeymasterSpec extends BorderPatrolSuite  {
+class KeymasterSpec extends BorderPatrolSuite with MockitoSugar {
```
trane: *sigh*. we knew it would happen, just didn't think it would be so soon :(

i wish i had more time to help you not do this. i'm sorry that i suck.

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -26,6 +26,7 @@ object Keymaster {
   case class KeymasterAccessReq(identity: Id[Tokens],
                                 serviceId: ServiceIdentifier, sessionId: SessionId) extends AccessRequest[Tokens]
   case class KeymasterAccessRes(access: Access[ServiceToken]) extends AccessResponse[ServiceToken]
+  case class KeymasterSessionIdRequest(req: SessionIdRequest, credential: Credential)
```
trane: so, we can't construct a `KeymasterIdentifyReq` because we need access to the `SessionId`? if so, why not just add that to the existing `IdentifyReq` on line `22`?

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -119,20 +119,20 @@ case class BorderService(identityProviderMap: Map[String, Service[SessionIdReque
     }
   }
 
-  def redirectTo(path: Path): Response =
-    tap(Response(Status.Found))(res => res.location = path.toString)
+  def redirectTo(pathStr: String): Response =
```
trane: why not make this `Path` and then call `.toString` in line `123`?

auth/src/test/scala/com/lookout/borderpatrol/auth/keymaster/KeymasterSpec.scala
maheshkelkar: Hrm. whats the alternative for using Mock

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
maheshkelkar: I changed it to `location`. Because it can be a URL or Path.

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
maheshkelkar: done

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
maheshkelkar: Yes that was the idea, last time around. Now, I moved everywhere to this Filter. 

And all transforms are moved to different methods in this filter. Note that oAuth2 stuff & proto manager is in `core`, Credentials (output) & this filter is in `auth`. Hence it was easier to do it this way.

trane: cool, so that makes the next version of this much easier - the type class version! 

the basic idea is we flip the dependency around, where we tell the compiler: "hey, do you have an instance of X that can transform this request into a Keymaster Credential?"

this is good for now, but let's plan on some time this week working on the other version. i think you'll find it nicer. :)

core/src/main/scala/com/lookout/borderpatrol/auth/OAuth2.scala
```diff
@@ -0,0 +1,174 @@
...
+    private[this] def download_aad_certs(downloadUrl: URL, thumbprint: String): Future[String] = {
+      //  Fetch the response
+      BinderBase.connect("AADCert-" + downloadUrl.toString, Set(downloadUrl),
+        Request(downloadUrl.getPath)).flatMap(res => res.status match {
+
+        //  Parse for Tokens if Status.Ok
+        case Status.Ok => {
+
+          // Load xml
+          val xml = Try(scala.xml.XML.loadString(res.contentString.replaceAll("[^\\x20-\\x7e]", ""))) match {
```
trane: why not wrap this in the try future wrapper thing?

PR #112, opened 2016-01-12
BP should match both subdomain and path to find Service

### LINKED ISSUE #111: Border Patrol service matcher should look at both subdomain and path
Currently, BP filters by subdomain and then by path. Thus, may choose an incorrect ServiceIdentifier.

Instead, it should look at combination of subdomain & path to determine the serviceIdentifier, so that following config can work:

subdomain: foo, path: /fun
subdomain: foo, path: /sun

maheshkelkar:
- Fix support for multiple services with same subdomain
- Added CustomerIdentifier to represent customer specific config

Fixes #111

### DIFF
--- a/bpConfig.json
+++ b/bpConfig.json
@@ -54,18 +54,26 @@
   "serviceIdentifiers": [
     {
       "hosts": ["http://localhost:8081"],
-      "loginManager": "checkpoint",
-      "subdomain": "ent",
-      "path": "/ent",
+      "path": "/serv",
       "name": "one"
     },
     {
       "hosts": ["http://localhost:8081"],
-      "loginManager": "dilli",
-      "subdomain": "dilli",
       "path": "/top",
       "rewritePath": "/internal/top",
-      "name": "dilli"
+      "name": "two"
+    }
+  ],
+  "customerIdentifiers": [
+    {
+      "loginManager": "checkpoint",
+      "subdomain": "ent",
+      "defaultServiceIdentifier": "one"
+    },
+    {
+      "loginManager": "dilli",
+      "subdomain": "api",
+      "defaultServiceIdentifier": "two"
     }
   ]
 }
--- a/core/src/main/scala/com/lookout/borderpatrol/ServiceMatcher.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/ServiceMatcher.scala
@@ -78,11 +78,15 @@ case class ServiceMatcher(customerIds: Set[CustomerIdentifier], serviceIds: Set[
 
   /**
    * Derive a CustomerIdentifier and ServiceIdentifier from an `httpx.Request`
+   * - Find CustomerIdentifier from `subdomain` from `req.host`
+   * - Find ServiceIdentifier from `req.path`.
+   *   If it fails to find, then check if it matches with paths in LoginManager. If it does,
+   *   then use default ServiceIdentifier
    */
-  def get(req: Request): Option[Tuple2[CustomerIdentifier, ServiceIdentifier]] =
+  def get(req: Request): Option[(CustomerIdentifier, ServiceIdentifier)] =
     (req.host.flatMap(subdomain), path(Path(req.path))) match {
-      case (Some(cid), Some(sid)) => Some(Tuple2(cid, sid))
-      case (Some(cid), None) if (cid.isLoginManagerPath(Path(req.path))) => Some(Tuple2(cid, cid.defaultServiceId))
+      case (Some(cid), Some(sid)) => Some((cid, sid))
+      case (Some(cid), None) if (cid.isLoginManagerPath(Path(req.path))) => Some((cid, cid.defaultServiceId))
       case _ => None
     }
 }
--- a/core/src/main/scala/com/lookout/borderpatrol/ServiceMatcher.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/ServiceMatcher.scala
@@ -4,38 +4,37 @@ import com.twitter.finagle.httpx.Request
 import com.twitter.finagle.httpx.path.Path
 
 /*
- * We derive a service `name` (a [[String]] name referencing a [[com.twitter.finagle.Name Name]]) either via the `Path`
- * or given a path of `/` we use the subdomain of the host.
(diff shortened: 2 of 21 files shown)


bpConfig.json
```diff
@@ -54,18 +54,26 @@
...
       "hosts": ["http://localhost:8081"],
-      "loginManager": "dilli",
-      "subdomain": "dilli",
       "path": "/top",
       "rewritePath": "/internal/top",
-      "name": "dilli"
+      "name": "two"
+    }
+  ],
+  "customerIdentifiers": [
```
trane: curious, would it be helpful to have a `name`?

core/src/main/scala/com/lookout/borderpatrol/ServiceMatcher.scala
```diff
@@ -50,28 +49,41 @@ case class ServiceMatcher(services: Set[ServiceIdentifier]) {
...
 
   /**
-   * Derive a ServiceIdentifier from an `httpx.Request`
+   * Derive a CustomerIdentifier and ServiceIdentifier from an `httpx.Request`
    */
-  def get(req: Request): Option[ServiceIdentifier] =
-    req.host.flatMap(subdomain).filter(sid => sid.isMatchingPath(Path(req.path)))
+  def get(req: Request): Option[Tuple2[CustomerIdentifier, ServiceIdentifier]] =
+    (req.host.flatMap(subdomain), path(Path(req.path))) match {
+      case (Some(cid), Some(sid)) => Some(Tuple2(cid, sid))
```
trane: you don't need the `Tuple2` you can just do `Some((cid, sid))`

core/src/main/scala/com/lookout/borderpatrol/ServiceMatcher.scala
```diff
@@ -50,28 +49,41 @@ case class ServiceMatcher(services: Set[ServiceIdentifier]) {
...
+        if (sid1.path.toString.size > sid2.path.toString.size) sid1 else sid2
     )
 
   /**
-   * Derive a ServiceIdentifier from an `httpx.Request`
+   * Derive a CustomerIdentifier and ServiceIdentifier from an `httpx.Request`
    */
-  def get(req: Request): Option[ServiceIdentifier] =
-    req.host.flatMap(subdomain).filter(sid => sid.isMatchingPath(Path(req.path)))
+  def get(req: Request): Option[Tuple2[CustomerIdentifier, ServiceIdentifier]] =
```
trane: `def get(req: Request): Option[(CustomerIdentifier, ServiceIdentifier)]`

core/src/main/scala/com/lookout/borderpatrol/ServiceMatcher.scala
```diff
@@ -50,28 +49,41 @@ case class ServiceMatcher(services: Set[ServiceIdentifier]) {
...
+  def path(path: Path): Option[ServiceIdentifier] =
+    foldWith(
+      serviceIds.filter(sid => path.startsWith(sid.path)),
+      (sid1: ServiceIdentifier, sid2: ServiceIdentifier) =>
+        if (sid1.path.toString.size > sid2.path.toString.size) sid1 else sid2
     )
 
   /**
-   * Derive a ServiceIdentifier from an `httpx.Request`
+   * Derive a CustomerIdentifier and ServiceIdentifier from an `httpx.Request`
```
trane: describe the 3 cases here

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -17,10 +17,10 @@ import scala.util.{Failure, Success}
 /**
  * PODs
  */
-case class ServiceRequest(req: Request, serviceId: ServiceIdentifier)
+case class ServiceRequest(req: Request, customerId: CustomerIdentifier, serviceId: ServiceIdentifier)
```
trane: does it even make sense to have this distinction between `ServiceRequest` and `SessionIdRequest`? especially with that overloaded constructor on line 22.

we can remove some of those nested `req.req.req` calls if we re-signature these:

```
case class ServiceRequest(req: Request, customerId: CustomerIdentifier, serviceId: ServiceIdentifier)

case class SessionIdRequest(req: Request, customerId: CustomerIdentifier, serviceId: ServiceIdentifier, sessionId: SessionId) {
  def apply(sr: ServiceRequest, sid: SessionId): SessionIdRequest =
    this(sr.req, sr.customerId, sr.serviceId, sessionId)
}

case class AccessIdRequest[A](req: Request, customerId: CustomerIdentifier, serviceId: ServiceIdentifier, sessionId: SessionId, id: Id[A]) {
  def apply(sr: SessionIdRequest, id: Id[A]): AccessIdRequest[A] =
    this(sr.req, sr.customerId, sr.serviceId, sr.sessionId, id)
}
```

bpConfig.json
maheshkelkar: no. We don't need `name`, unless of course if customer wants name to be different than subdomain.

core/src/main/scala/com/lookout/borderpatrol/ServiceMatcher.scala
maheshkelkar: Compile complained when I did that.

core/src/main/scala/com/lookout/borderpatrol/ServiceMatcher.scala
maheshkelkar: compiler complained. But, will try again

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
maheshkelkar: ok. done.

core/src/main/scala/com/lookout/borderpatrol/ServiceMatcher.scala
maheshkelkar: done

maheshkelkar:
@trane Let me know if you are all set with this and I will merge it. Thanks

PR #113, opened 2016-01-15
ConsulStore improvements

### LINKED ISSUE #90: Consul Backed Secret Store Improvements
Consul backed Secret Store has a few places where it can be improved.

- Look into rewriting `set` function in the `ConsulConnection` class so does not need a url. Removing this can also remove the need to pass host and port to different classes
- Consolidating classes in SecretStore and Secrets Cache. 
- Removing the ConsulConnection class and using an httpx.Service or Client. Move the functionality of value into the place where responses are used.
- Rewrite the tests to not depend on `sleep`
- Create an abstract and reusable SecretsCache. Possibly something similar to this https://gist.github.com/tejom/1c86eb0e2f8497bb958b
- @trane I think has plans to remove ` object ConsulSecretStore ` and the constructors

maheshkelkar:
- Added cas/modifyIndex based get-put-get mechanism to manage Secrets on Consul
- Consolidated SecretCache (ConsulSecretCache, ConsulConnection) into ConsulSecretStore
- Using Circe for decoding ConsulResponse
- Added unit tests for the code coverage
- If Secrets are expired, then deny the login

Fixes #90


core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -1,11 +1,25 @@
 package com.lookout.borderpatrol.sessionx
 
-
 import java.net.URL
+import java.util.concurrent.TimeUnit
+import java.util.logging.Logger
+import javax.xml.bind.DatatypeConverter
+
+import argonaut.Parse
```
trane: why do we use both argonaut and circe? can we just kill argonaut completely and use the circe+jawn parser please?

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -60,84 +74,218 @@ object SecretStores {
...
+   * @param consulUrl
+   * @param _secrets
    */
-  case class ConsulSecretStore(consul:  ConsulConnection, poll: Long, secretsStart: Secrets) extends SecretStoreApi {
-    val cache = ConsulSecretCache(poll, consul, secretsStart)
-    new Thread(cache).start()
+  case class ConsulSecretStore(consulUrl: URL, var _secrets: Secrets)
+      extends SecretStoreApi {
+    private[this] val log = Logger.getLogger(getClass.getName)
+    private[this] val ConsulSecretsKey = "/v1/kv/BpSecrets"
```
trane: let's make this configurable via Config and constructor.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -60,84 +74,218 @@ object SecretStores {
...
+
+    /* Alt constructor: this is the commonly used constructor. The main constructor is merely used for unit testing */
+    def this(consulUrl: URL) = this(consulUrl, Secrets(Secret(Time.now), Secret(Time.now)))
```
trane: can we make this `apply` instead in a companion object? having to call `new` is not very idiomatic and makes for more clunky apis.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -60,84 +74,218 @@ object SecretStores {
...
-   * @param consul An instance of [[com.lookout.borderpatrol.sessionx.ConsulConnection]] to make requests
-   *               to the consul server
-   * @param poll How often in seconds to check for updated Secrets on the consul server
+   * @param consulUrl
+   * @param _secrets
    */
-  case class ConsulSecretStore(consul:  ConsulConnection, poll: Long, secretsStart: Secrets) extends SecretStoreApi {
-    val cache = ConsulSecretCache(poll, consul, secretsStart)
-    new Thread(cache).start()
+  case class ConsulSecretStore(consulUrl: URL, var _secrets: Secrets)
```
trane: let's make `_secrets` an immutable stream or an append-log instead of a mutable `var`? that's the idea behind the secret cache, iirc -- but that was made with a mutable array.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -60,84 +74,218 @@ object SecretStores {
...
+            _secrets = s
+            DefaultTimer.twitter.schedule(_secrets.current.expiry)(pollSecrets)
+
+          case _ =>
+            log.log(Level.DEBUG, "ConsulSecretStore: Failed to get Secrets from Consul, trying again soon in 1 minute")
+            DefaultTimer.twitter.schedule(Time.now + Duration(1, TimeUnit.MINUTES))(pollSecrets)
+
+        }) handle {
+        case e: Throwable =>
+          log.log(Level.ERROR, s"ConsulSecretStore: Failed to get Secrets from Consul with: ${e.getMessage}, " +
```
trane: we'll definitely want to hook up twitter stats to this so we can be alerted through stats, e.g. datadog

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
maheshkelkar: Secrets/SessionStore etc are still using Argonaut. I decided not to touch it right now. I will do it later.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
maheshkelkar: I assume `/v1/kv` is fixed. Only thing configurable is BpSecrets. Since this is an internal thing, why make it configurable?

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -60,84 +74,218 @@ object SecretStores {
...
-      val consulConnection = new ConsulConnection(client, consulUrl.getHost, consulUrl.getPort.toString)
-      val secretsDefault = Secrets(Secret(Time.fromMilliseconds(0)), Secret())
-      new ConsulSecretStore(consulConnection, poll, secretsDefault)
-    }
+    def pollSecrets(): Unit =
+      (for {
+        (consulResponse, secrets, result) <- getSecretsFromConsul(1, false, None)
+        (secrets, result) <- putSecretsOnConsul(2, result, consulResponse, secrets)
+        (_, secrets, result) <- getSecretsFromConsul(3, result, secrets)
+      } yield (secrets, result) match {
```
trane: let's put this match into it's own function and `yield funcName(secrets, result)`

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -60,84 +74,218 @@ object SecretStores {
...
+        Future[(Option[ConsulResponse], Option[Secrets], Boolean)] =
+      /* Skip this step, if result is already successful */
+      if (result) {
+        (None, inputSecrets, true).toFuture
+      } else {
+
+        /* Get the secrets from Consul */
+        for {
+          consulResponseList <- BinderBase.connect(consulUrl.toString, Set(consulUrl),
+                                                   Request(ConsulSecretsKey)).flatMap(
```
trane: let's put this in its own function too and have it return a `Future...`

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -60,84 +74,218 @@ object SecretStores {
...
+                  t => Future.value(t))
+
+              case Status.NotFound => List.empty.toFuture
+
+              case _ => Future.exception(ConsulError(
+                s"Step-${step}: Failed with an error response from Consul: ${res.status}"))
+
+            })
+          consulResponse <- consulResponseList.headOption.toFuture
+          secrets <- consulResponse match {
```
trane: let's pull this match into it own function too

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
maheshkelkar: Yes it was a mutable array. But, why store it in the array? If we do then we have manage the contents of array as well?
Or do we always maintain 1 `Secrets` in the array

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -60,84 +74,218 @@ object SecretStores {
...
+            case None => None.toFuture
+            case Some(cr) => Some(cr.secretsForValue()).toFuture
+          }
+          result <- secrets.fold(false)(!_.current.expired).toFuture
```
trane: why fold on secrets instead of just `(! secrets.current.expired).toFuture`?

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
maheshkelkar: I tried, but, I couldnt pass statsReceiver to this, but we create statsReceiver in different location, made it dififcult for me to attach it here.

Ler me keep looking into this

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
maheshkelkar: I thought about it, but then it may seem like we are doing same type of things in 2 different places.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -60,84 +74,218 @@ object SecretStores {
...
+                                         inputSecrets: Option[Secrets]):
+        Future[(Option[Secrets], Boolean)] =
+      /* Skip this step, if result is already successful */
+      if (result) {
+        (inputSecrets, true).toFuture
+      } else {
+
+        /* Generate next secrets */
+        val newSecrets = inputSecrets match {
+          case None => Secrets(Secret(), _secrets.current)
```
trane: i think the following algorithm makes sense to create the new secret with the following expiry:

_secrets.current.expiry + 24 hours.

Then it will be more deterministic in when we resolve to update the secrets.

Thoughts?

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
maheshkelkar: I had actually done that; but then restored it as it seem only 2 lines of code

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -60,84 +74,218 @@ object SecretStores {
...
+        val modifyIndex = inputConsulResponse match {
+          case None => 0
+          case Some(oc) => oc.modifyIndex
+        }
+
+        /* Try setting the next secrets */
+        BinderBase.connect(consulUrl.toString, Set(consulUrl),
+          tap(Request(Method.Put, Request.queryString(ConsulSecretsKey, ("cas" -> modifyIndex.toString))))(req => {
+            req.contentString = SecretsEncoder.EncodeJson.encode(newSecrets).toString
+            req.contentType = "text/plain"
```
trane: isn't the contentType `"application/json"`?

maheshkelkar: yes. fixed it

core/src/main/scala/com/lookout/borderpatrol/sessionx/Session.scala
```diff
@@ -80,10 +82,15 @@ object Session {
...
    * @return Session
    */
-  def apply[A](data: A, tag: Tag = Untagged)(implicit store: SecretStoreApi): Future[Session[A]] = tag match {
-    case AuthenticatedTag => SessionId.authenticated map (Session(_, data))
-    case Untagged => SessionId.untagged map (Session (_, data) )
-    case _ => new SessionError("Invalid SessionId Tag found").toFutureException
-  }
+  def apply[A](data: A, tag: Tag = Untagged)(implicit store: SecretStoreApi): Future[Session[A]] =
+    if (store.current.expired) {
+        new SessionError("Logins are currently down, sorry for the inconvenience").toFutureException
```
trane: The error text shouldn't be in the exception -- keep in mind that any errors that make it to a user should be translated into their language.
This should be a new type of `SessionError`... one that can be sent to whatever displays the error pages in the correctly translated text. Maybe `SessionCreateUnavailable` or something.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
trane: we will be storing all the secrets we've seen since BP started in there for the same reasons that you would use an append-only-log instead of mutating something in-place.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
trane: what if you are using the same consul cluster with multiple border patrol clusters? you would be forcing the same secrets to be used by all of the different border patrol clusters.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
trane: we can attach stats to logging

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
trane: i'm only asking for the `match` statement to be its own function -- this function is too big to understand quickly.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
trane: i mean the `Request(...).flatMap` thing should go in its own function... so it makes this `for` comprehension look like this:
```
for {
  consulResponseList <- BinderBase.connect(consulUrl.toString, Set(consulUrl), blahblahFuncName(ConsulSecretsKey))
  consulResponse <- consulResponseList.headOption.toFuture
  secrets <- funcName(consulResponse)
  result <- secrets.fold(false)!(_current.expired).toFuture
} yield (consulResponse, secrets, result)
```

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
maheshkelkar: ok, makes sense. will change it

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
maheshkelkar: done

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
maheshkelkar: will do a separate issue.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
maheshkelkar: Hrm, not sure I understand how we can do that. Are you suggesting we create a wrapper? But, we need to create statsReceiver in advance for that to work. Lets discuss this.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
maheshkelkar: done

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
maheshkelkar: done

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
maheshkelkar: done

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
maheshkelkar: done

maheshkelkar:
@trane updated after code review comments.

core/src/main/scala/com/lookout/borderpatrol/sessionx/Errors.scala
```diff
@@ -25,3 +25,6 @@ case class OriginalRequestNotFound(msg: String)
 
 case class ConsulError(msg: String)
   extends SessionError(s"An error occurred getting a value from Consul: $msg")
+
+case class SessionCreateUnavailable()
+  extends SessionError(s"Logins are currently down, sorry for the inconvenience")
```
trane: no, the session error message is not for users, it's for us. it should say why. "only expired secrets available" or something.

PR #115, opened 2016-01-16
Add logout support

### LINKED ISSUE #114: Add a support for /logout to delete the session
and redirect the use back to default service

maheshkelkar:
- simply listens for logouts on a hardcoded path `/logout`
- if Request contains the cookie for `SessionId`, then the Session is looked up and removed from the SessionStore
- The client is redirected to default Service Id dictated by the subdomain (i.e. configured in customerIdentifier)

Fixes #114

### DIFF
--- a/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
@@ -168,22 +168,15 @@ case class LogoutService(store: SessionStore)(implicit secretStore: SecretStoreA
   private[this] val log = Logger.getLogger(getClass.getSimpleName)
 
   def apply(req: ServiceRequest): Future[Response] = {
-    SessionId.fromRequest(req.req) match {
-      case Success(sid) => {
-        log.log(Level.DEBUG, s"Logging out Session: ${sid.toLogIdString}")
-        store.delete(sid)
-        tap(Response(Status.Found)) { res =>
-          res.location = req.customerId.defaultServiceId.path.toString
-          res.addCookie(SessionId.toCookie(sid, true))
-          log.log(Level.DEBUG, s"W/ Session: Redirecting to default service path: ${res.location}")
-        }.toFuture
-      }
-      case Failure(e) =>
-        tap(Response(Status.Found)) { res =>
-          res.location = req.customerId.defaultServiceId.path.toString
-          log.log(Level.DEBUG, s"W/O Session: Redirecting to default service path: ${res.location}")
-        }.toFuture
-    }
+    SessionId.fromRequest(req.req).foreach(sid => {
+      log.log(Level.DEBUG, s"Logging out Session: ${sid.toLogIdString}")
+      store.delete(sid)
+    })
+    tap(Response(Status.Found)) { res =>
+      res.location = req.customerId.defaultServiceId.path.toString
+      res.addCookie(SessionId.toExpiredCookie())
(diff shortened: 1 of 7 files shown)


core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -158,6 +158,36 @@ case class BorderService(identityProviderMap: Map[String, Service[SessionIdReque
...
+  private[this] val log = Logger.getLogger(getClass.getSimpleName)
+
+  def apply(req: ServiceRequest): Future[Response] = {
+    SessionId.fromRequest(req.req) match {
+      case Success(sid) => {
+        log.log(Level.DEBUG, s"Logging out Session: ${sid.toLogIdString}")
+        store.delete(sid)
+        tap(Response(Status.Found)) { res =>
+          res.location = req.customerId.defaultServiceId.path.toString
+          res.addCookie(SessionId.toCookie(sid, true))
```
trane: Shouldn't this be remove?

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -158,6 +158,36 @@ case class BorderService(identityProviderMap: Map[String, Service[SessionIdReque
...
+          log.log(Level.DEBUG, s"W/ Session: Redirecting to default service path: ${res.location}")
+        }.toFuture
+      }
+      case Failure(e) =>
+        tap(Response(Status.Found)) { res =>
```
trane: This is exactly the same as above save the store delete,  dry?

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
trane: Hrm, I see. I'm really not a fan of this boolean carrying the concept of an expired cookie

maheshkelkar: `removeCookie` simply removes a cookie of that name from Response.cookieMap. So we don't want that.
A couple of options are: 
- reset maxAge or 
- set the isDiscard flag in the cookie or 
- clear cookie value

We are setting isDicard flag

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
maheshkelkar: Yes, the code to generate response is same, but log message and `store.delete` is different. 

Let me think of a way to do the store.delete if its a success.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SessionId.scala
```diff
@@ -101,6 +101,14 @@ object SessionId {
       cookie.maxAge = Time.now.until(sessionId.expires)
     }
 
+  def toExpiredCookie(): Cookie =
+    tap(new Cookie("border_session", "")) { cookie =>
```
trane: this is my fault, i realize, but let's make `"border_session"` a `val` on the `SessionId` object and use the reference in `toCookie` and `toExpiredCookie`

maheshkelkar: Its used in Encoder.scala as well. Let me take a look at how we can remove it

maheshkelkar:
@trane done. let me know if you are ok with this. If not, i will merge it

PR #117, opened 2016-01-18
Add CSRF insert filter and filters to chain

### LINKED ISSUE #116: Add CSRF to the filter chain
Insert CSRF filters in the chain such a way that they can actively insert and verify the cookie

maheshkelkar:
Fixes #116


security/src/test/scala/com/lookout/borderpatrol/test/security/CsrfSpec.scala
```diff
@@ -16,6 +16,20 @@ class CsrfSpec extends BorderPatrolSuite {
...
   val verify = Verify(InHeader(header), Param(param), CookieName(cookiename), VerifiedHeader(verified))
 
+  val filter = CsrfVerifyFilter(verify)
+
+  val service = Service.mk[Request, Response]{ req =>
+    Future.value {
+      req.headerMap.get(verified) match {
+        case Some("true") => Response(Status.Ok)
+        case Some("false") => Response(Status.Forbidden)
+        case _ => throw new Exception("bottom")
```
trane: this should be an exception that we control that inherits from `Exception`

server/src/main/scala/com/lookout/borderpatrol/server/services.scala
```diff
@@ -11,6 +16,32 @@ import com.twitter.finagle.stats.StatsReceiver
 
 object services {
 
+
+  /**
+   *  Keymaster Identity provider service Chain
+   * @param store
+   */
+  def keymasterIdentityProviderChain(store: SessionStore)(
```
trane: none of these `keymaster` things belong in the `server` module since `keymaster` is not open sourced.

server/src/test/scala/com/lookout/borderpatrol/server/ServicesSpec.scala
```diff
@@ -2,37 +2,159 @@ package com.lookout.borderpatrol.server
...
+import com.twitter.finagle.httpx.{Status, Response, Request}
 import com.twitter.finagle.httpx.path.Path
+import com.twitter.util.Await
 
 class ServicesSpec extends BorderPatrolSuite {
   import sessionx.helpers._
   import services._
 
-  val urls = Set(new URL("http://localhost:5678"))
+//  val urls = Set(new URL("http://localhost:5678"))
```
trane: hrm, lots of comments here. just remove them if you aren't going to use it. it looks like they are still in version control from before.

maheshkelkar: Sorry, I was supposed to delete this code. But, forgot

server/src/main/scala/com/lookout/borderpatrol/server/services.scala
maheshkelkar: The `keymaster` stuff is just a guidance on how we can build a chain together. It used to be int he `auth`, now its here becasue of `CSRF` filter. I didnt want to add `security` module dependency in `auth` module.

`server` puts all of these things together.  `keymaster` being part of the `auth` module is already open source. Are you suggesting we remove `keymaster` from `auth` module?

maheshkelkar: @trane do you want me to move this object to `example`

trane: if that makes more sense than in `auth.keymaster` sure.

maheshkelkar: I can restore and move it back to `auth.keymaster`. We cannot add `CsrfInsertFilter` there without adding `auth` depend on `security`. Please note that we have to insert `CsrfInsertFilter` in the path which is travelled only once after successful authentication, and hence I decided to put it there.

So if we need to use CSRF, we need to create a seperate keymaster identity service chain in the BP internal repository.

trane: CsrfInsertFilter should be able to be pluggable wherever it should be needed. The fact that it goes into the Keymaster stuff for an example app is not a problem.

our internal repo is a great place to generate a service chain with keymaster included!

maheshkelkar:
@trane abandoning this, I will submit a new one

PR #118, opened 2016-01-20
Add CSRF insert filter

### LINKED ISSUE #116: Add CSRF to the filter chain

maheshkelkar:
- also moved MainServiceChain to example
- clear all border_ cookies on logout

Fixes #116


core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -174,7 +175,16 @@ case class LogoutService(store: SessionStore)(implicit secretStore: SecretStoreA
     })
     tap(Response(Status.Found)) { res =>
       res.location = req.customerId.defaultServiceId.path.toString
-      res.addCookie(SessionId.toExpiredCookie())
+      req.req.cookies.foreach[Unit] {
+        case (name: String, cookie: Cookie) if name.startsWith("border_") =>
+          // Expire all BP cookies
+          res.addCookie(
```
trane: Don't we have a method called toExpiredCookie?

core/src/main/scala/com/lookout/borderpatrol/sessionx/SessionId.scala
```diff
@@ -102,14 +102,6 @@ object SessionId {
       cookie.maxAge = Time.now.until(sessionId.expires)
     }
 
-  def toExpiredCookie(): Cookie =
```
trane: Oh, I guess not anymore

security/src/main/scala/com/lookout/borderpatrol/security/Csrf.scala
```diff
@@ -48,10 +48,32 @@ object Csrf {
...
+ * - It should be typically happen only once, perhaps after the successful login
+ */
+case class CsrfInsertFilter[A](cookieName: Csrf.CookieName)(implicit secretStore: SecretStoreApi)
+    extends Filter[A, Response, A, Response] {
+
+  def apply(req: A, service: Service[A, Response]): Future[Response] =
+    service(req).flatMap(res =>
+      for {
+        csrfSessionId <- SessionId.authenticated
+        cookie <- tap(new Cookie(cookieName.name, csrfSessionId.asBase64)) { cooki => {
```
trane: Can't toCookie take an argument for the cookie name so it can be used here?

maheshkelkar: I thought about it; but then I thought `border_session` & `SessionId` are closely interlinked. 

In general I wasn't sure.

trane: `SessionId` is the perfect place for it, since we are using that logic to generate a signed cookie. We could rename `SessionId` to `SignedId` or something like that to ensure that we are communicating what its purpose is.

On a side note:

Names are hard, because what we want is the type of `(Timestamp, Entropy, Signature)` for many things. If you think of the type, it will be clearer what should be used where. The more we can rely on the type of a thing the more likely we are to compose our code correctly and reuse elements and find generic uses. Names are hard, and honestly `SessionId` isn't a great name because it's a use for a timestamped-cryptographically-signed thing.

maheshkelkar: Changing SessionId -> SignedId, means changing 371 occurrences (over 30-40 files)

trane: I'm not saying you should. But, keep in mind that our code is statically compiled - which means that we can have confidence in our IDE to do renaming without breaking our code.

PR #121, opened 2016-01-25
After login return 404 for unknown resources

### LINKED ISSUE #120: After login BP should return 404 for unknown resources and redirect to login page before login
* Post login, BP should return 404 for unknown resources
* Pre login, BP should redirect requests (with valid or invalid paths) to login page
* And in all cases should redirect Root path to default Service

maheshkelkar:
* Post login, BP should return 404 for unknown resources
* Pre login, BP should redirect requests (with valid or invalid paths) to login page
* And in all cases should redirect Root path to default Service

Fixes #120

### DIFF
--- a/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
@@ -17,22 +17,22 @@ import scala.util.{Failure, Success}
 /**
  * PODs
  */
-case class ServiceRequest(req: Request, customerId: CustomerIdentifier)
-case class BorderRequest(req: Request, customerId: CustomerIdentifier, sessionId: SignedId)
-object BorderRequest {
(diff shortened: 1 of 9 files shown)


core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -150,11 +150,18 @@ case class BorderService(identityProviderMap: Map[String, Service[SessionIdReque
   }
 
   def apply(req: SessionIdRequest): Future[Response] =
```
trane: should this param be turned into a `BorderRequest`?

maheshkelkar: `BorderService` takes `SessionIdRequest` as input and pushes `BorderRequest` to next (follow-on) chains.

trane:
lgtm, feel free to merge once travis comes back successful

PR #124, opened 2016-01-27
Add support for unprotected upstream services

### LINKED ISSUE #123: Add support for unprotected upstream services
Thus, BorderService categorizes requests as follows:
* Request to LoginConfirm path via IdentityProvider chain
* Request to unprotected upstream service, directly
* Request to protected upstream service via AccessIssuer chain

It simplifies the login manager config:
* redirect to internal path for internal authentication
* redirect to external URL for external/SSO authentication

Additional things:
* That helps us make Checkpoint an "unprotected" service. Also, using `rewritePath` feature in `ServiceIdentifier` we can keep the externally visible and internally mounted paths different.
* As shown in the example, it allows us to make logout path configurable (currently hardcoded as `/logout`), which in turn can be serviced by internally routed Logout Service

maheshkelkar:
- Checkpoint, ErrorRedirect, logout all become unprotected upstream services
- Remove LoginManagerBinder (it doesn’t connect to upstream anymore)
- Remove LoginManagerFilter (replaced by unprotectedServiceChain inside BorderService)
- Moved alternate constructor for StatsdExporter to object (so that we can avoid the new)

Also:
- Fixed afterEach to call super
- Fixed ConsulSecretStore to use HashWheeledTimer instead of DefaultTimer.twitter (deprecated)
- Fixed SecretStoreSpec to close the timer
- Cleaned up BinderSpec

Fixes #123

### DIFF
--- a/bpConfig.json
+++ b/bpConfig.json
@@ -22,8 +22,7 @@
       "proto": {
         "type": "Internal",
         "loginConfirm": "/signin",
-        "path": "/a",
-        "hosts": ["http://localhost:8081"]
+        "authorizePath": "/a"
       }
     },
     {
@@ -64,6 +63,19 @@
       "path": "/top",
       "rewritePath": "/internal/top",
       "name": "two"
+    },
+    {
+      "hosts": ["http://localhost:8081"],
+      "path": "/a",
+      "protected": false,
+      "name": "checkpoint"
+    },
+    {
+      "hosts": ["http://localhost:8081"],
+      "path": "/logout",
+      "rewritePath": "/signout",
+      "protected": false,
+      "name": "logout"
     }
   ],
   "customerIdentifiers": [
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,7 @@
 import sbtunidoc.Plugin.UnidocKeys._
 import scoverage.ScoverageSbtPlugin.ScoverageKeys.coverageExcludedPackages
 
-lazy val Version = "0.1.7-SNAPSHOT"
+lazy val Version = "0.1.8-SNAPSHOT"
 
 lazy val buildSettings = Seq(
   organization := "com.lookout",
--- a/core/src/main/scala/com/lookout/borderpatrol/Manager.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/Manager.scala
@@ -3,6 +3,7 @@ package com.lookout.borderpatrol
 import java.net.URL
 import com.twitter.finagle.http.{Method, Response, Request}
 import com.twitter.finagle.http.path.Path
+import com.twitter.logging.Logger
 import com.twitter.util.Future
 
 case class Manager(name: String, path: Path, hosts: Set[URL])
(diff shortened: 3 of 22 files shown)


bpConfig.json
```diff
@@ -64,6 +63,19 @@
       "path": "/top",
       "rewritePath": "/internal/top",
       "name": "two"
+    },
+    {
+      "hosts": ["http://localhost:8081"],
+      "path": "/a",
+      "protected": false,
```
trane: just to be clear... the design is to create a service identifier for *every* unprotected path?

why not make a value of `unprotectedPaths` of type `List[Path]`? that way you can have a "whitelist" per service identifier?

build.sbt
```diff
@@ -1,7 +1,7 @@
 import sbtunidoc.Plugin.UnidocKeys._
 import scoverage.ScoverageSbtPlugin.ScoverageKeys.coverageExcludedPackages
 
-lazy val Version = "0.1.7-SNAPSHOT"
+lazy val Version = "0.1.8-SNAPSHOT"
```
PeterFCaswell: Just curious: why is this lazy? It's not like it's an assignment that's going to take up a lot of CPU cycles that you want to avoid.

core/src/main/scala/com/lookout/borderpatrol/Manager.scala
```diff
@@ -3,6 +3,7 @@ package com.lookout.borderpatrol
 import java.net.URL
 import com.twitter.finagle.http.{Method, Response, Request}
 import com.twitter.finagle.http.path.Path
+import com.twitter.logging.Logger
 import com.twitter.util.Future
 
 case class Manager(name: String, path: Path, hosts: Set[URL])
```
PeterFCaswell: Could we get some comments here telling us what the class does?  What is it managing?

PeterFCaswell: Could we get some comments here telling us what the class does? What is it managing?

core/src/main/scala/com/lookout/borderpatrol/Manager.scala
```diff
@@ -13,30 +14,24 @@ case class LoginManager(name: String, identityManager: Manager, accessManager: M
 trait ProtoManager {
```
PeterFCaswell: Is this a "protocol" manager? Or a "prototype" manager? Or some other proto?
Comments are wonderful things. :-)

core/src/main/scala/com/lookout/borderpatrol/ServiceIdentifier.scala
```diff
@@ -14,7 +14,8 @@ import com.twitter.finagle.http.path.Path
  * @param rewritePath The (optional) internal url path prefix to the internal service. If present,
  *                    it replaces the external path in the Request URI
  */
```
PeterFCaswell: Need a comment here for the 'protekted' parameter. Is this with funny spelling because 'protected' is a keyword?

build.sbt
maheshkelkar: Typically, lazy vals are used instead of vals to avoid initialization order problems.

PeterFCaswell: Okay. But this is a straight constant. I don't understand how there could be order problems.

It doesn't really matter. It's fine.

core/src/main/scala/com/lookout/borderpatrol/Manager.scala
maheshkelkar: done

core/src/main/scala/com/lookout/borderpatrol/Manager.scala
maheshkelkar: done

core/src/main/scala/com/lookout/borderpatrol/ServiceIdentifier.scala
maheshkelkar: yes. done

bpConfig.json
maheshkelkar: Yes, current design is to create a service identifier for every unprotected path. In future, we have talked about extending `path: Path` variable to `paths: List[Path]`, which will extend it to multiple unprotected paths on a given set of hosts

I thought about your suggestion for adding a `unprotectedPaths: List[Path]` instead of `protected: Boolean`. I thought about some obvious issues such as (how do we decide default unprotected path, rewritePaths, what about serviceIdentifier without any protected paths, etc) and yes we can work around all of these issues. 

So, this is mainly a configuration issue, currently its designed as:
```
"ServiceIdentifiers" : [ 
// protected
{ "hosts": ["http://localhost:8080"], "path": "/good", "name": "good" }, 
// protected w/ rewrite
{ "hosts": ["http://localhost:8081"], "path": "/out", rewritePath: "/in", "name": "rewrite" }, 
// unprotected (on same host)
{ "hosts": ["http://localhost:8081"], "path": "/open", "protected": false, "name": "open" }
// unprotected w/ rewrite
{ "hosts": ["http://localhost:8082"], "path": "/another", "rewritePath": "/what", "protected": false, "name": "another" }
]
```
"ServiceIdentifiers" : [ 
// protected (no unprotected paths)
{ "hosts": ["http://localhost:8080"], "path": "/good", "unprotected": [], "name": "good" },
// protected w/ rewrite & 1 unprotected path
{ "hosts": ["http://localhost:8081"], "path": "/out", rewritePath: "/in", "unprotected": ["/out"], "name": "rewrite" },
// hack protected & 1 unprotected path
{ "hosts": ["http://localhost:8081"], "path": "/notfound", "unprotected": ["/another"], "name": "open" }
]
```

PR #125, opened 2016-02-02
Add config for port, external endpoint, loggedOut url

maheshkelkar:
port - port where each BP node anchors
external endpoint - visible to outside work; E.g. ELB address (and port if necessary)
loggedOut url - after a logout, user is redirected to this URL (in order to add a landing area after SSO logout)

### DIFF
--- a/bpConfig.json
+++ b/bpConfig.json
@@ -1,4 +1,6 @@
 {
+  "port": "8080",
+  "externalEndpoint": "http://localhost:8080",
   "statsdReporter": {
      "host": "localhost:8125",
      "durationInSec": 60,
@@ -35,6 +37,7 @@
         "authorizeUrl": "https://login.zoo.com:443/common/oauth2/authorize",
         "tokenUrl": "https://login.zoo.com:443/common/oauth2/token",
         "certificateUrl": "https://login.zoo.com:443/common/certificate",
+        "loggedOutUrl": "http://www.zoo.com",
         "clientId": "MySecretClientId",
         "clientSecret": "MySecretClientSecret"
       }
--- a/core/src/main/scala/com/lookout/borderpatrol/Manager.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/Manager.scala
@@ -31,7 +31,8 @@ case class LoginManager(name: String, identityManager: Manager, accessManager: M
  */
 trait ProtoManager {
   val loginConfirm: Path
-  def redirectLocation(host: Option[String]): String
+  val loggedOutUrl: Option[URL]
+  def redirectLocation: String
   def isMatchingPath(p: Path): Boolean = Set(loginConfirm).filter(p.startsWith(_)).nonEmpty
 }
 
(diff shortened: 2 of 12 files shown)


bpConfig.json
```diff
@@ -1,4 +1,6 @@
 {
+  "port": "8080",
+  "externalEndpoint": "http://localhost:8080",
```
PeterFCaswell: Is the 8080 here redundant? Why have the port listed separately?

core/src/main/scala/com/lookout/borderpatrol/Manager.scala
```diff
@@ -41,38 +42,42 @@ trait ProtoManager {
...
  * @param authorizeUrl URL of the OAuth2 service where client is redirected for authenticaiton
  * @param tokenUrl URL of the OAuth2 server to convert OAuth2 code to OAuth2 token
  * @param certificateUrl URL of the OAuth2 server to fetch the certificate for verifying token signature
+ * @param loggedOutUrl A Url where user is redirected after the Logout
  * @param clientId Id used for communicating with OAuth2 server
  * @param clientSecret Secret used for communicating with OAuth2 server
  */
-case class OAuth2CodeProtoManager(loginConfirm: Path, authorizeUrl: URL, tokenUrl: URL, certificateUrl: URL,
-                                  clientId: String, clientSecret: String)
+case class OAuth2CodeProtoManager(bpExternalEndpoint: URL, loginConfirm: Path, authorizeUrl: URL, tokenUrl: URL,
```
PeterFCaswell: I find it's not a particularly good idea to have a series of arguments of the same type. There's no clue about whether you've put them in the right order. I'd strongly encourage you to make it a TODO or an enhancement Jira to yourself to find some way of disambiguating between these arguments. I'd be happy to talk about ways I've dealt with that in the past.

core/src/main/scala/com/lookout/borderpatrol/Manager.scala
```diff
@@ -41,38 +42,42 @@ trait ProtoManager {
...
-  def redirectLocation(host: Option[String]): String = {
-    val hostStr = host.getOrElse(throw new Exception("Host not found in HTTP Request"))
+  def redirectLocation: String =
     Request.queryString(authorizeUrl.toString, ("response_type", "code"), ("state", "foo"),
-      ("client_id", clientId), ("redirect_uri", "http://" + hostStr + loginConfirm.toString))
-  }
-  def codeToToken(host: Option[String], code: String): Future[Response] = {
-    val hostStr = host.getOrElse(throw new Exception("Host not found in HTTP Request"))
+      ("client_id", clientId),
+      ("redirect_uri", s"$bpExternalEndpoint$loginConfirm"))
```
PeterFCaswell: This seems really strange. Composing two URL's together doesn't seem like something useful without something between them. Beyond that, do we have some validation that the two URLs are well formed and will mean something if they are composed like this?

core/src/main/scala/com/lookout/borderpatrol/Manager.scala
```diff
@@ -41,38 +42,42 @@ trait ProtoManager {
...
+
+  def codeToToken(code: String): Future[Response] = {
     val request = util.Combinators.tap(Request(Method.Post, tokenUrl.toString))(re => {
       re.contentType = "application/x-www-form-urlencoded"
       re.contentString = Request.queryString(("grant_type", "authorization_code"), ("client_id", clientId),
-        ("code", code), ("redirect_uri", "http://" + hostStr + loginConfirm.toString),
+        ("code", code),
+        ("redirect_uri", s"$bpExternalEndpoint$loginConfirm"),
```
PeterFCaswell: Same concern about composing the URLs.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -116,7 +116,8 @@ object SecretStores {
     private[this] def processPolledSecrets(secrets: Option[Secrets], result: Boolean): Unit =
       (secrets, result) match {
         case (Some(s), true) =>
-          log.info(s"ConsulSecretStore: Received a new Secret from Consul with an id: ${s.current.id}")
+          log.info("ConsulSecretStore: Received a new Secret from Consul " +
+            s"with an id: ${s.current.id}, expiry: ${s.current.expiry}")
```
PeterFCaswell: Given this form of string appears twice does it make sense to make it a method on Secrets to format the string "with and id..."?

bpConfig.json
maheshkelkar: 8080 is where borderpatrol `node` is listening. External endpoint is really an externally visible URL of the borderpatrol (so typically, ELB)

PeterFCaswell:
I don't see enough to block the pull, but it'll be good if you can look at the issues I've listed in a future PR.

core/src/main/scala/com/lookout/borderpatrol/Manager.scala
maheshkelkar: one of them is URL and other is path. I will add a method to compose them together

core/src/main/scala/com/lookout/borderpatrol/Manager.scala
maheshkelkar: done

PR #128, opened 2016-02-04
Show all config validation errors at once

### LINKED ISSUE #127: Show all config validation errors at once and fix the parsing error text
Currently, config validation stops at first error. instead we should display all the errors at once, for better experience.
Also, error messages are vague (i.e. failed to decode following fields: CustomerIdentifier). It should display actual error message that caused parsing failure (albeit in some cases).

maheshkelkar:
Currently, config validation stops at first error. Instead we should display
all the errors at once, for better experience.

Also, error messages are vague (i.e. failed to decode following fields:
CustomerIdentifier). It should display actual error message that caused the parsing
failure (albeit in some cases).

Fixes #127

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,7 @@
 import sbtunidoc.Plugin.UnidocKeys._
 import scoverage.ScoverageSbtPlugin.ScoverageKeys.coverageExcludedPackages
 
-lazy val Version = "0.1.10-SNAPSHOT"
+lazy val Version = "0.1.11-SNAPSHOT"
 
 lazy val buildSettings = Seq(
   organization := "com.lookout",
--- a/server/src/main/scala/com/lookout/borderpatrol/server/Config.scala
+++ b/server/src/main/scala/com/lookout/borderpatrol/server/Config.scala
@@ -7,8 +7,6 @@ import com.lookout.borderpatrol.sessionx.SecretStores._
 import com.lookout.borderpatrol.sessionx.SessionStores._
 import com.lookout.borderpatrol.sessionx._
 import com.twitter.finagle.Memcached
-import com.twitter.finagle.Name
-import com.twitter.finagle.Resolver
 import com.twitter.finagle.http.path.Path
 import com.twitter.app.App
 import cats.data.Xor
@@ -52,8 +50,7 @@ object Config {
   val defaultConfigFile = "bpConfig.json"
   val defaultSecretStore = SecretStores.InMemorySecretStore(Secrets(Secret(), Secret()))
   val defaultSessionStore = SessionStores.InMemoryStore
-  val serverConfigFields = Set("secretStore", "sessionStore", "customerIdentifiers", "serviceIdentifiers",
-    "loginManagers", "identityManagers", "accessManager", "statsdReporter", "listeningPort")
+  def cond[T](p: => Boolean, v: T) : Set[T] = if (p) Set(v) else Set.empty[T]
 
   // Encoder/Decoder for Path
   implicit val encodePath: Encoder[Path] = Encoder[String].contramap(_.toString)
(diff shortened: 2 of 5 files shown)


build.sbt
```diff
@@ -1,7 +1,7 @@
 import sbtunidoc.Plugin.UnidocKeys._
 import scoverage.ScoverageSbtPlugin.ScoverageKeys.coverageExcludedPackages
 
-lazy val Version = "0.1.10-SNAPSHOT"
+lazy val Version = "0.1.11-SNAPSHOT"
```
PeterFCaswell: Why is "SNAPSHOT" in the version string?

server/src/main/scala/com/lookout/borderpatrol/server/Config.scala
```diff
@@ -162,10 +159,10 @@ object Config {
         name <- c.downField("name").as[String]
         ipName <- c.downField("identityManager").as[String]
         im <- Xor.fromOption(ims.get(ipName),
-          DecodingFailure(s"No IdentityManager $ipName found: ", c.history))
+          DecodingFailure(s"IdentityManager - $ipName - not found: ", c.history))
```
PeterFCaswell: Why not put quotes around the name? Otherwise it's not clear whether the spaces are in the name or not?

server/src/main/scala/com/lookout/borderpatrol/server/Config.scala
```diff
@@ -162,10 +159,10 @@ object Config {
...
         apName <- c.downField("accessManager").as[String]
         am <- Xor.fromOption(ams.get(apName),
-          DecodingFailure(s"No AccessManager $apName found: ", c.history)
+          DecodingFailure(s"AccessManager - $apName - not found: ", c.  history)
```
PeterFCaswell: Here and below too.

build.sbt
maheshkelkar: I represents development release. We publish these to jfrog. The real releases goto bintray

server/src/main/scala/com/lookout/borderpatrol/server/Config.scala
maheshkelkar: ok

server/src/main/scala/com/lookout/borderpatrol/server/Config.scala
maheshkelkar: ok

server/src/main/scala/com/lookout/borderpatrol/server/Config.scala
```diff
@@ -338,24 +332,24 @@ object Config {
    *
    * @param serverConfig
    */
-  def validate(serverConfig: ServerConfig): Unit = {
+  def validate(serverConfig: ServerConfig): Set[String] = {
     //  Validate Secret Store config
-    validateSecretStoreConfig("secretStore", serverConfig.secretStore)
+    validateSecretStoreConfig("secretStore", serverConfig.secretStore) ++ (
```
PeterFCaswell: So after staring at this for a while, I came to the conclusion that we were concatenating sets of error messages from each validation step. Is that the case? Can you add a comment or two here?

PeterFCaswell: Also, why the parenthesis at the end of this line?

server/src/main/scala/com/lookout/borderpatrol/server/Config.scala
```diff
@@ -366,9 +360,16 @@ object Config {
    */
   def readServerConfig(filename: String) : ServerConfig = {
     decode[ServerConfig](Source.fromFile(filename).mkString) match {
```
PeterFCaswell: This could use some comments too. If I was deep in the code, would I know what the Xor here was doing? Is that a standard Scala-ism?

maheshkelkar: Xor is used by `Circe` our JSON parser

server/src/main/scala/com/lookout/borderpatrol/server/Config.scala
maheshkelkar: I will add a comment. The parenthesis allows me to concatenate Sets across multi-line.

PR #131, opened 2016-03-03
Split BorderService into seperate filters

### LINKED ISSUE #130: Split BorderService into multiple filters
Also, allow a POST to loginConfirm w/o session-id to continue with authentication without an extra 302 roundtrip. It helps in case of performance testing.

maheshkelkar:
- Split into IdentityService, AccessIssuer and UnprotectedService filters
- Allow POST to loginConfirm w/o sessionId to continue with authentication

Fixes #130

### DIFF
--- a/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
@@ -2,30 +2,34 @@ package com.lookout.borderpatrol.auth
 
 import com.lookout.borderpatrol.Binder.{BindRequest, MBinder}
 import com.lookout.borderpatrol.util.Combinators._
-import com.lookout.borderpatrol.{CustomerIdentifier, LoginManager, ServiceIdentifier, ServiceMatcher}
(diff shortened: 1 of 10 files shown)


core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -241,11 +301,10 @@ case class IdentityFilter[A : SessionDataEncoder](store: SessionStore)(implicit
     (for {
       sessionMaybe <- store.get[A](sessionId)
     } yield sessionMaybe.fold[Identity[A]](EmptyIdentity)(s => Id(s.data))) handle {
-      case e => {
+      case e =>
```
PeterFCaswell: Why did you remove the curly braces? Doesn't that change the code? Does EmptyIdentity only apply to the case e or to every case?

maheshkelkar: These braces are not needed.

PeterFCaswell: Not needed by the compiler, but maybe needed for human comprehension? I always find the scala terse-is-good and no-extra-characters paradigms to make it difficult to read. But whatever.

PR #132, opened 2016-03-07
Add support for application/JSON calls in BP2

maheshkelkar:
- Added prefix "Bp" to all the BP errors/exceptions
- Throw Bp exceptions for all the error responses generated by BP2
- Convert exceptions to error responses in the ExceptionFilter
- If request is exepcting an application/json response, then return
  a response with custom response (e.g. contains a redirect_url)

### DIFF
--- a/auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
+++ b/auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
@@ -1,6 +1,7 @@
 package com.lookout.borderpatrol.auth.keymaster
 
 import com.lookout.borderpatrol.auth.OAuth2.OAuth2CodeVerify
+import com.lookout.borderpatrol.errors.{ForbiddenRequest, BadRequest}
 import com.lookout.borderpatrol.util.Combinators.tap
 import com.lookout.borderpatrol.sessionx._
 import com.lookout.borderpatrol._
@@ -45,6 +46,8 @@ object Keymaster {
       statsReceiver.counter("keymaster.identity.provider.response.success")
     private[this] val responseFailed =
       statsReceiver.counter("keymaster.identity.provider.response.failed")
+    private[this] val responseDenied =
+      statsReceiver.counter("keymaster.identity.provider.denied")
 
     /**
      * Sends credentials, if authenticated successfully will return a MasterToken otherwise a Future.exception
@@ -60,21 +63,22 @@ object Keymaster {
           Tokens.derive[Tokens](res.contentString).fold[Future[IdentifyResponse[Tokens]]](
             err => {
               responseParsingFailed.incr
-              Future.exception(IdentityProviderError(Status.InternalServerError,
-                "Failed to parse the Keymaster Identity Response"))
+              Future.exception(BpTokenParsingError("Failed to parse the Keymaster Identity Response"))
             },
             t => {
               responseSuccess.incr
               Future.value(KeymasterIdentifyRes(t))
             }
           )
-        //  Preserve Response Status code by throwing AccessDenied exceptions
+        case Status.Forbidden => {
+          responseDenied.incr
+          Future.exception(BpIdentityProviderError(Status.Forbidden,
+            s"IdentityProvider denied user: ${req.credential.uniqueId} with status: ${res.status}"))
+        }
         case _ => {
-          log.debug(s"Keymaster IdentityProvider denied user: ${req.credential.uniqueId} " +
-            s"with status: ${res.status}")
           responseFailed.incr
-          Future.exception(IdentityProviderError(res.status,
-            s"Invalid credentials for user ${req.credential.uniqueId}"))
+          Future.exception(BpIdentityProviderError(Status.InternalServerError,
+            s"IdentityProvider denied user: ${req.credential.uniqueId} with status: ${res.status}"))
         }
       })
     }
@@ -92,8 +96,7 @@ object Keymaster {
         p <- req.req.params.get("password")
       } yield InternalAuthCredential(u, p, req.customerId, req.serviceId)) match {
         case Some(c) => Future.value(c)
-        case None => Future.exception(IdentityProviderError(Status.InternalServerError,
-          "transformBasic: Failed to parse the Request"))
+        case None => Future.exception(new BadRequest("Failed to find username and/or password in the Request"))
       }
 
     def transformOAuth2(req: BorderRequest, protoManager: OAuth2CodeProtoManager): Future[OAuth2CodeCredential] = {
@@ -135,7 +138,7 @@ object Keymaster {
     def requestFromSessionStore(sessionId: SignedId): Future[Request] =
       store.get[Request](sessionId).flatMap {
         case Some(session) => Future.value(session.data)
-        case None => Future.exception(OriginalRequestNotFound(s"no request stored for ${sessionId.toLogIdString}"))
+        case None => Future.exception(BpOriginalRequestNotFound(s"no request stored for ${sessionId.toLogIdString}"))
       }
 
     def apply(req: KeymasterIdentifyReq,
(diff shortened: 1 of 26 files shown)


auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -60,21 +63,22 @@ object Keymaster {
...
-        //  Preserve Response Status code by throwing AccessDenied exceptions
+        case Status.Forbidden => {
+          responseDenied.incr
+          Future.exception(BpIdentityProviderError(Status.Forbidden,
+            s"IdentityProvider denied user: ${req.credential.uniqueId} with status: ${res.status}"))
+        }
         case _ => {
-          log.debug(s"Keymaster IdentityProvider denied user: ${req.credential.uniqueId} " +
-            s"with status: ${res.status}")
           responseFailed.incr
```
PeterFCaswell: Would it be clearer here if the variable was named, for instance "statResponseFailed"? This applies to the other statistic names.

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -60,21 +63,22 @@ object Keymaster {
           Tokens.derive[Tokens](res.contentString).fold[Future[IdentifyResponse[Tokens]]](
             err => {
               responseParsingFailed.incr
-              Future.exception(IdentityProviderError(Status.InternalServerError,
-                "Failed to parse the Keymaster Identity Response"))
+              Future.exception(BpTokenParsingError("Failed to parse the Keymaster Identity Response"))
```
PeterFCaswell: Can you explain why you've added the 'Bp' prefix? Aren't all these names already qualified with a borderpatrol namespace?

core/src/main/scala/com/lookout/borderpatrol/Binder.scala
```diff
@@ -100,11 +100,9 @@ object BinderBase {
       cl <- getOrCreate(name, urls)
       res <- cl.apply(request)
     } yield res) handle {
-      case e => {
-        log.warning("Failed to connect " +
-          s"for: $name to: ${urls.map(u => u.getAuthority).mkString(",")} with: ${e.getMessage}")
-        throw CommunicationError(s"${name} with ${e.getMessage}")
-      }
+      case e =>
+        throw BpCommunicationError(s"Failed to connect for: $name " +
```
PeterFCaswell: I would recommend quotes around things like $name. It is a pain in the ass to be reading an error message where parameters like this have spaces in them and you can't easily determine what is the name and what is the error message.

(And no, I don't always remember to do this, and yes, please point it out to me when I miss it. :-) )

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -370,28 +340,75 @@ case class RewriteFilter() extends SimpleFilter[BorderRequest, Response] {
...
-    case error: IdentityProviderError => warningAndResponse(error.getMessage, error.status)
-    case error: Exception =>
-      log.error(error, error.getMessage)
-      tap(Response(Status.InternalServerError))(
-        r => { r.contentString = error.getMessage; r.contentType = "text/plain" }
-      )
+  def errorHandler(req: Request): PartialFunction[Throwable, Response] = {
+    case error: BpAccessIssuerError => warningAndResponse(req, error.getMessage, error.status)
+    case error: BpIdentityProviderError => warningAndResponse(req, error.getMessage, error.status)
+    case error: BpRedirectError =>
```
PeterFCaswell: A comment here would be famous.

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -370,28 +340,75 @@ case class RewriteFilter() extends SimpleFilter[BorderRequest, Response] {
...
+              ("redirect_url", error.location.asJson)
+            )).toString()
+            res.contentType = "application/json"
+          case _ =>
+            res.status = Status.Found
+            res.location = error.location
+            res.contentString = error.msg
+            res.contentType = "text/plain"
+        }})
+    case error: BpLogoutError =>
```
PeterFCaswell: A comment here would be much appreciated.

core/src/main/scala/com/lookout/borderpatrol/package.scala
```diff
@@ -56,8 +56,9 @@ package object borderpatrol {
   }
 
   object errors {
-    abstract class BorderError(val status: http.Status, val description: String) extends Exception
-    class InvalidRequest(description: String = "") extends BorderError(http.Status.BadRequest, description)
+    class BorderError(val status: http.Status, val description: String) extends Exception(description)
```
PeterFCaswell: If you're prefixing 'Bp' to all your errors, shouldn't these have the same prefix?

core/src/main/scala/com/lookout/borderpatrol/sessionx/SignedId.scala
```diff
@@ -156,7 +156,7 @@ object SignedId {
     implicit def bytes2Secret(bytes: IndexedSeq[Byte])(implicit store: SecretStoreApi): Try[Secret] =
       store.find(_.id == bytes) match {
         case Some(s) => Success(s)
-        case None => Failure(new SignedIdError(s"No secret with id=$bytes"))
+        case None => Failure(new BpSignedIdError(s"No secret with id=$bytes"))
```
PeterFCaswell: This is another case where delimiters around $bytes would be really helpful to human readers. Especially if the bytes randomly have newlines in them.

PeterFCaswell:
That's all I have for this go-round.

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
maheshkelkar: ok, will fix it.

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
maheshkelkar: Almost all the Borderpatrol Error(s) are far too generic (e.g. CommunicationError, SessionError, etc) and hard to differentiate from finagle or java errors/exceptions. Hence I decided to add a `Bp` prefix

core/src/main/scala/com/lookout/borderpatrol/Binder.scala
maheshkelkar: ok, will do.

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
maheshkelkar: done.

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
maheshkelkar: done :)

core/src/main/scala/com/lookout/borderpatrol/package.scala
maheshkelkar: Yes

core/src/main/scala/com/lookout/borderpatrol/sessionx/SignedId.scala
maheshkelkar: ok

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -370,28 +339,82 @@ case class RewriteFilter() extends SimpleFilter[BorderRequest, Response] {
...
+          res.status = Status.Found
+          res.location = error.location
+          res.contentString = error.msg
+          res.contentType = "text/plain"
+      }})
+  }
+
+  /* Convert this error into a response appropriate to the client */
+  private[this] def warningAndLogoutResponse(req: Request, error: BpLogoutError): Response = {
+    log.info(error.msg)
```
PeterFCaswell: Why is this at info and the log above is at warning level?

PeterFCaswell:
Just a minor issue. If you say the log level difference is on purpose, I'll merge it.

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
maheshkelkar: I should change the method name.

Logout & Redirect, both should be at info level. I will update it.

PR #137, opened 2016-03-29
Updated docs and minor fixes

maheshkelkar:
- Updated README.md with Block Diagram, Concepts, Components,
  Configuration, etc
- Updated contents of some error logs
- Added code to scrub special characters from query parameters
- Added logs to dump X-Forwarded-* for debugging
- Fix the version in HealthCheck output


auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -73,12 +74,12 @@ object Keymaster {
         case Status.Forbidden => {
           statResponseDenied.incr
           Future.exception(BpIdentityProviderError(Status.Forbidden,
-            s"IdentityProvider denied user: '${req.credential.uniqueId}' with status: ${res.status}"))
+            s"IdentityProvider failed to authenticate user: '${req.credential.uniqueId}'"))
```
PeterFCaswell: Does req.credential.uniqueId contain the username? If so, you need to redact that.

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -73,12 +74,12 @@ object Keymaster {
...
         }
         case _ => {
           statResponseFailed.incr
           Future.exception(BpIdentityProviderError(Status.InternalServerError,
-            s"IdentityProvider denied user: '${req.credential.uniqueId}' with status: ${res.status}"))
+            s"IdentityProvider failed to authenticate user: '${req.credential.uniqueId}', with status: ${res.status}"))
```
PeterFCaswell: Same as above.

PeterFCaswell:
Your CI failed. :-( 
But the above are the only issues that I could see.

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
maheshkelkar: Hi Peter, it indeed contains a username. We have a couple of more log messages that dump sensitive information such as token or username, x-forwarded-for. I will take them off as soon as we are done testing in the staging.

PR #139, opened 2016-03-30
Ensure default serviceId for a CustomerIdentifier is protected

maheshkelkar:
- Use ParamMap instead of req.uri, since it fetches query param
  from uri (GET) or contentString (POST)
- Log an error message if we see multiple ServiceIdentifiers
  for a tuple of (service name and protected flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ libraryDependencies ++= Seq(
 
 ```scala
 libraryDependencies ++= Seq(
-  "com.lookout.borderpatrol" %% "[borderpatrol-module]" % "0.1.18-SNAPSHOT"
+  "com.lookout.borderpatrol" %% "[borderpatrol-module]" % "0.1.20-SNAPSHOT"
 )
 ```
 
@@ -108,8 +108,7 @@ Configuration
 -------------
 
  * `secretStore`: Secret Store. It can be configured using `type` as `InMemorySecretStore` or `ConsulSecretStore`.
-   * `InMemorySecretStore`: Typically used for single host setup as Secrets are meant to be shared across all the
-BorderPatrol nodes
+   * `InMemorySecretStore`: Typically used for single host setup as Secrets are meant to be shared across all the BorderPatrol nodes
      ```json
      "secretStore" : {
        "type" : "InMemorySecretStore",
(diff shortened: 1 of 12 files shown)


maheshkelkar:
@pfc please hold off this review. I will let you know when its ready

maheshkelkar:
Abandoning, will submit a new one

README.md
```diff
@@ -44,7 +44,7 @@ libraryDependencies ++= Seq(
 
 ```scala
 libraryDependencies ++= Seq(
-  "com.lookout.borderpatrol" %% "[borderpatrol-module]" % "0.1.18-SNAPSHOT"
+  "com.lookout.borderpatrol" %% "[borderpatrol-module]" % "0.1.20-SNAPSHOT"
```
PeterFCaswell: What happened to 0.1.19?

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -91,14 +91,15 @@ object Keymaster {
   case class KeymasterTransformFilter(oAuth2CodeVerify: OAuth2CodeVerify)(implicit statsReceiver: StatsReceiver)
       extends Filter[BorderRequest, Response, KeymasterIdentifyReq, Response] {
 
-    def transformInternal(req: BorderRequest): Future[InternalAuthCredential] =
+    def transformInternal(req: BorderRequest): Future[InternalAuthCredential] = {
       (for {
-        u <- Helpers.scrubQueryParams(req.req.uri, "username")
-        p <- Helpers.scrubQueryParams(req.req.uri, "password")
+        u <- Helpers.scrubQueryParams(req.req.params, "username")
```
PeterFCaswell: Not something to delay this PR for, but would it be cleaner if you defined a method in BorderRequest, called "params", that simply invoked "req.params"?

core/src/main/scala/com/lookout/borderpatrol/util/Helpers.scala
```diff
@@ -1,35 +1,20 @@
...
-import org.jboss.netty.handler.codec.http.QueryStringDecoder
+import com.twitter.finagle.http.ParamMap
 
-import scala.collection.JavaConverters._
-import scala.util.{Success, Try}
 
 object Helpers {
 
-  /** Parse uri into (path, params) */
+  /** Regular Expression for all special characters from ASCII 0 to 20 */
```
PeterFCaswell: Is this Octal 0-20 or decimal 0-20? If the latter, what about the characters from NAK to SP?

PR #140, opened 2016-03-30
Use Request ParamMap instead of req.uri

maheshkelkar:
- since it fetches query param from uri (GET) or contentString (POST)

Also,
- Log an error message if we see multiple ServiceIdentifiers for a
  tuple of (service name and protected flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ libraryDependencies ++= Seq(
 
 ```scala
 libraryDependencies ++= Seq(
-  "com.lookout.borderpatrol" %% "[borderpatrol-module]" % "0.1.18-SNAPSHOT"
+  "com.lookout.borderpatrol" %% "[borderpatrol-module]" % "0.1.20-SNAPSHOT"
 )
 ```
 
@@ -108,8 +108,7 @@ Configuration
 -------------
 
  * `secretStore`: Secret Store. It can be configured using `type` as `InMemorySecretStore` or `ConsulSecretStore`.
-   * `InMemorySecretStore`: Typically used for single host setup as Secrets are meant to be shared across all the
-BorderPatrol nodes
+   * `InMemorySecretStore`: Typically used for single host setup as Secrets are meant to be shared across all the BorderPatrol nodes
      ```json
      "secretStore" : {
        "type" : "InMemorySecretStore",
@@ -126,8 +125,7 @@ BorderPatrol nodes
      }
      ```
  * `sessionStore`: Session Store. It can be configured using `type` as `InMemoryStore` or `MemcachedStore`.
-   * `InMemoryStore`: Typically used for single host setup as Sessions are meant to be shared across all the
-BorderPatrol nodes.
+   * `InMemoryStore`: Typically used for single host setup as Sessions are meant to be shared across all the BorderPatrol nodes.
      ```json
      "sessionStore" : {
        "type" : "InMemoryStore",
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -44,7 +44,7 @@ libraryDependencies ++= Seq(
 
 ```scala
 libraryDependencies ++= Seq(
-  "com.lookout.borderpatrol" %% "[borderpatrol-module]" % "0.1.18-SNAPSHOT"
+  "com.lookout.borderpatrol" %% "[borderpatrol-module]" % "0.1.20-SNAPSHOT"
```
PeterFCaswell: What happened to 0.1.19?

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -91,14 +91,15 @@ object Keymaster {
   case class KeymasterTransformFilter(oAuth2CodeVerify: OAuth2CodeVerify)(implicit statsReceiver: StatsReceiver)
       extends Filter[BorderRequest, Response, KeymasterIdentifyReq, Response] {
 
-    def transformInternal(req: BorderRequest): Future[InternalAuthCredential] =
+    def transformInternal(req: BorderRequest): Future[InternalAuthCredential] = {
       (for {
-        u <- Helpers.scrubQueryParams(req.req.uri, "username")
-        p <- Helpers.scrubQueryParams(req.req.uri, "password")
+        u <- Helpers.scrubQueryParams(req.req.params, "username")
```
PeterFCaswell: Not something to delay this PR for, but would it be cleaner if you defined a method in BorderRequest, called "params", that simply invoked "req.params"?

core/src/main/scala/com/lookout/borderpatrol/util/Helpers.scala
```diff
@@ -1,35 +1,20 @@
...
-import org.jboss.netty.handler.codec.http.QueryStringDecoder
+import com.twitter.finagle.http.ParamMap
 
-import scala.collection.JavaConverters._
-import scala.util.{Success, Try}
 
 object Helpers {
 
-  /** Parse uri into (path, params) */
+  /** Regular Expression for all special characters from ASCII 0 to 20 */
```
PeterFCaswell: Is this Octal 0-20 or decimal 0-20? If the latter, what about the characters from NAK to SP?

I really think you need octal 20 or decimal 32.

security/src/main/scala/com/lookout/borderpatrol/security/Csrf.scala
```diff
@@ -40,7 +40,7 @@ object Csrf {
      */
     def verify(req: Request): Boolean =
       (for {
-        str <- req.headerMap.get(header.header) orElse Helpers.scrubQueryParams(req.uri, param.param)
+        str <- req.headerMap.get(header.header) orElse Helpers.scrubQueryParams(req.params, param.param)
```
PeterFCaswell: "param.param" seems really awkward. Does Param ever have any value other than "_x_border_csrf"?  Wouldn't this be cleaner as "x_crsf.val"?

This won't block the PR, but you should consider it for the future.

PeterFCaswell:
Only things preventing the merge is the special characters you're filtering and the CI build failure.

README.md
maheshkelkar: Forgot to update README.md

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
maheshkelkar: I will add it to the list. Thanks

core/src/main/scala/com/lookout/borderpatrol/util/Helpers.scala
maheshkelkar: I am sorry, thats supposed to be hex. I will correct it

security/src/main/scala/com/lookout/borderpatrol/security/Csrf.scala
maheshkelkar: I will fix that.

maheshkelkar:
@pfc I fixed the comments, and CI is passing now

PR #143, opened 2016-04-12
Change return type for CsrfInsertFilter

### LINKED ISSUE #142: Use a case class for CsrfInsertFilter response
Currently, CsrfInsertFilter simply returns the `Response` object. Instead use a special pod "RedirectResponse" so that CSRF cookie can be appended before throwing a redirect exception

maheshkelkar:
- Previously, it was using "Response". Changing to an internal POD.
- BP is changed to throw an exception for all the internally generated
  errors or status codes; so that a proper response is returned in the
  ExceptionFilter
- Also updated README.md

Fixes #142

### DIFF
--- a/auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
+++ b/auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
@@ -86,10 +86,10 @@ object Keymaster {
   }
 
   /**
-   * Handles Keymaster transforms for internal and OAuth2
+   * Handles Keymaster transforms (incoming request into Keymaster specific request) for internal and OAuth2
    */
   case class KeymasterTransformFilter(oAuth2CodeVerify: OAuth2CodeVerify)(implicit statsReceiver: StatsReceiver)
-      extends Filter[BorderRequest, Response, KeymasterIdentifyReq, Response] {
+      extends Filter[BorderRequest, Response, KeymasterIdentifyReq, RedirectResponse] {
 
     def transformInternal(req: BorderRequest): Future[InternalAuthCredential] = {
       (for {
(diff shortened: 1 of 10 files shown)


auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -109,14 +109,14 @@ object Keymaster {
...
+              service: Service[KeymasterIdentifyReq, RedirectResponse]): Future[Response] = {
       for {
         transformed: Credential <- req.customerId.loginManager.protoManager match {
           case a: InternalAuthProtoManager => transformInternal(req)
           case b: OAuth2CodeProtoManager => transformOAuth2(req, b)
         }
-        resp <- service(KeymasterIdentifyReq(req, transformed))
-      } yield resp
+        redirectResp <- service(KeymasterIdentifyReq(req, transformed))
+      } yield throw BpRedirectError(redirectResp.status, redirectResp.location, redirectResp.cookies, redirectResp.msg)
```
mbbx6spp: Since the default case is to throw an error (`BpRedirectError`) can we make it part of the return value instead of throwing it? I am not sure a redirect should be considered "exceptional" and I would like to remove as many exceptions being used for the purpose of control flow by the application itself.

PeterFCaswell:
I too would love to see the exceptions removed for normal cases.

Other than that, I don't have any comments. Won't merge until hearing what you think about that issue.

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
maheshkelkar: Thanks for the review Susan.

In general, I am treating any response generated by BP as an exception:
* The normal return type is `http.Response` and Status.302 for redirect; if client expects JSON, then its needs to be formatted differently. So by throwing exception, it allows me to format it properly in one place. 

Let me re-evaluate it.


-

maheshkelkar:
@mbbx6spp @PeterFCaswell Thanks for your comments. Abandoning. this and submitting a new pull request

PR #145, opened 2016-04-25
Cleanup logs and prep for v0.2.0 release

maheshkelkar:
- Also, do not return error message back to caller
- Use X-Forwarded-Proto to determine if retdirect url should be
  http or https

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,7 @@
 import sbtunidoc.Plugin.UnidocKeys._
 import scoverage.ScoverageSbtPlugin.ScoverageKeys.coverageExcludedPackages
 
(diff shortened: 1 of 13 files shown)


build.sbt
```diff
@@ -1,7 +1,7 @@
 import sbtunidoc.Plugin.UnidocKeys._
 import scoverage.ScoverageSbtPlugin.ScoverageKeys.coverageExcludedPackages
 
-lazy val Version = "0.1.22-SNAPSHOT"
+lazy val Version = "0.2.0"
```
PeterFCaswell: Whoot! 0.2.0!

PR #146, opened 2016-04-25
Split BP config, so that app can compose it as per the requirements

maheshkelkar:
- Split the config such a way that `server` module contains all the composable encoder and decoders for the pieces (such as `CustomerIdentifier`, `ServiceIdentifier`, `LoginManager`, etc)
- The example module can composed together in a ServerConfig. And in future can it into into AppConfig and DataConfig, etc
- Removed `Manager`(s) and replaced with `Endpoint`(s).
- `Endpoint` represents a REST endpoint remote to BP (i.e. combination of Set(urls) and path) where BP can send a query for execution.
- The `Endpoint` can be used to configure `LoginManger` and `HealthCheckEndpoints`
- The key difference between `ServiceIdentifier` and `Endpoint` is BP hosts/represents/maps the `ServiceIdentifier` path prefix on the subdomain of the `CustomerIdentifier`. So any requests hitting the combination of subdomain+path is represented by BP. Whereas `Endpoint`s enable BP to achieve that.
- Updated documentation
- Added 2 login managers `keymaster.basic` and `keymaster.oauth2`


auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -57,7 +54,7 @@ object Keymaster {
       statRequestSends.incr
 
       //  Authenticate user by the Keymaster
-      binder(BindRequest(req.credential.customerId.loginManager.identityManager, req.credential.toRequest))
+      Binder.connect(req.credential.customerId.loginManager.identityEndpoint, req.credential.toRequest)
```
PeterFCaswell: Is this using the Binder class's companion class object (the singleton)? Is that really better than passing an object in as a parameter?

auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/Keymaster.scala
```diff
@@ -101,19 +98,19 @@ object Keymaster {
       }
     }
 
-    def transformOAuth2(req: BorderRequest, protoManager: OAuth2CodeProtoManager): Future[OAuth2CodeCredential] = {
+    def transformOAuth2(req: BorderRequest, loginManager: OAuth2LoginManager): Future[OAuth2CodeCredential] = {
```
PeterFCaswell: I like 'loginManager' much better than 'protoManager'!

core/src/main/scala/com/lookout/borderpatrol/ServiceIdentifier.scala
```diff
@@ -29,4 +30,5 @@ case class ServiceIdentifier(name: String, hosts: Set[URL], path: Path, rewriteP
  * @param defaultServiceId
  * @param loginManager
  */
-case class CustomerIdentifier(subdomain: String, defaultServiceId: ServiceIdentifier, loginManager: LoginManager)
+case class CustomerIdentifier(subdomain: String, guid: String, defaultServiceId: ServiceIdentifier,
```
PeterFCaswell: Perhaps you might have individual classes for "a string that contains a subdomain name" and "a string that contains the ASCII representation of a GUID", so you don't accidentally call this with the parameters in the wrong order.

Do you always treat the GUID as a string, and never its binary equivalent? (Just curious.)
This isn't enough for me to refuse the merge. But I strongly suggest you consider that in the future. (I wish I could do something like that in Ruby :-) )

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -170,11 +170,10 @@ case class SendToIdentityProvider(identityProviderMap: Map[String, Service[Borde
   def sendToIdentityProvider(req: BorderRequest): Future[Response] = {
     log.debug(s"Send: ${req.req} for Session: ${req.sessionId.toLogIdString} " +
       s"to identity provider chain for service: ${req.serviceId.name}")
-    identityProviderMap.get(req.customerId.loginManager.identityManager.name) match {
+    identityProviderMap.get(req.customerId.loginManager.ty) match {
```
PeterFCaswell: what does "ty" stand for?

example/src/main/scala/com/lookout/borderpatrol/example/service.scala
```diff
@@ -167,19 +170,18 @@ object service {
...
 
   // Mock Routing service
-  def getMockRoutingService(implicit config: ServerConfig, statsReceiver: StatsReceiver):
+  def getMockRoutingService(implicit config: Config, statsReceiver: StatsReceiver, secretStore: SecretStoreApi):
   Service[Request, Response] = {
-    val checkpoint = config.findServiceIdentifier("checkpoint")
-    val keymasterIdManager = config.findIdentityManager("keymaster")
-    val keymasterAccessManager = config.findAccessManager("keymaster")
+    val login = config.findServiceIdentifier("login")
+    val keymasterIdEndpoint = config.findEndpoint("keymaster-identity-example")
```
PeterFCaswell: I'd really like it if we'd remove the word "keymaster" from our public repo. I would think it would be a clearer example calling it "PasswordIdManager" or something like that. If we implement in our private repo talking to Keymaster that'd be cool.

For some later PR though.

server/src/main/scala/com/lookout/borderpatrol/server/Config.scala
```diff
@@ -283,18 +225,18 @@ object Config {
...
+   * Validate Endpoint configuration
    * @param field
-   * @param managers
+   * @param endpoints
    * @return set of all the errors encountered during validation
    */
-  def validateManagerConfig(field: String, managers: Set[Manager]): Set[String] = {
-    // Find if managers have duplicate entries
-    (cond(managers.size > managers.map(m => m.name).size,
+  def validateEndpointConfig(field: String, endpoints: Set[Endpoint]): Set[String] = {
```
PeterFCaswell: Would it be difficult to provide a way to run the validate scripts without spinning up BP? Some little utility program you could include with the repo? That would help us avoid deploy-time errors?

(Again, future PR)

PeterFCaswell:
Mahesh, consider the comments as conversation openers. If you think they're legit, please consider them for future PRs.

PR #147, opened 2016-04-26
Move keymaster login managers to auth module

maheshkelkar:
The login manager defines the identity and access management policy. Currently, we only have 2 policies defined in auth module, hence login managers belong in there.


auth/src/main/scala/com/lookout/borderpatrol/auth/keymaster/OAuth2.scala
```diff
@@ -148,11 +150,11 @@ object OAuth2 {
           case Status.Ok =>
             OAuth2.derive[AadToken](res.contentString).fold[Future[AadToken]](
               err => Future.exception(BpTokenParsingError(
-                s"in the Access Token response from OAuth2 Server: ${req.customerId.loginManager.name}")),
+                s"in the Access Token response from OAuth2 Server: ${loginManager.name}")),
```
PeterFCaswell: If the loginManager.name is an IdP name from KM, then it could have spaces in it, and I recommend you surround it by quotes. Nothing drives me crazier than reading a log file and realizing there's a trailing quote in a string that isn't supposed to be there. :-)

core/src/main/scala/com/lookout/borderpatrol/Manager.scala
```diff
@@ -21,72 +20,10 @@ case class Endpoint(name: String, path: Path, hosts: Set[URL])
  */
 trait LoginManager {
   val name: String
-  val ty: String
+  val tyfe: String
```
PeterFCaswell: What's a "tyfe"?

core/src/main/scala/com/lookout/borderpatrol/Manager.scala
```diff
@@ -21,72 +20,10 @@ case class Endpoint(name: String, path: Path, hosts: Set[URL])
...
   val guid: String
   val loginConfirm: Path
```
PeterFCaswell: What's the difference between a Path (which is the loginConfirm) and an Endpoint (the next two variables)?

PeterFCaswell:
Just a couple of nits. Nothing to block the merge.

PR #149, opened 2016-04-28
Replace keymaster with tokenmaster, an appropriate name

maheshkelkar:


PeterFCaswell:
Feel free to merge this when the checks pass.

maheshkelkar:
@PeterFCaswell It seems that codecov coverage went below 90%, raising some alarms. Since then I have added some more tests. But, I am still hitting the failure and I have not able to figure out problem. I have reached out to codecov folks and but it will take some time.

But, Travis is happy. IMO we should go ahead and merge this. Are you ok with this approach?

PR #154, opened 2016-05-05
User errors during authentication, redirect back to login page

### LINKED ISSUE #153: On any authentication failure, BorderPatrol should redirect user back to login page
And the redirect location would have query param "msg" with error string

maheshkelkar:
- anything else return 500

Fixes #153

### DIFF
--- a/auth/src/main/scala/com/lookout/borderpatrol/auth/tokenmaster/LoginManagers.scala
+++ b/auth/src/main/scala/com/lookout/borderpatrol/auth/tokenmaster/LoginManagers.scala
@@ -1,9 +1,10 @@
 package com.lookout.borderpatrol.auth.tokenmaster
 
+import com.lookout.borderpatrol.auth.BpInvalidRequest
 import com.lookout.borderpatrol.{Binder, Endpoint, LoginManager}
 import com.lookout.borderpatrol.util.Helpers
 import com.lookout.borderpatrol.util.Combinators.tap
-import com.twitter.finagle.http.{Method, Response, Request}
+import com.twitter.finagle.http.{Method, Request, Response}
 import com.twitter.finagle.http.path.Path
 import com.twitter.logging.Logger
 import com.twitter.util.Future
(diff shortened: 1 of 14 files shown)


auth/src/main/scala/com/lookout/borderpatrol/auth/tokenmaster/LoginManagers.scala
```diff
@@ -36,23 +38,24 @@ object LoginManagers {
     val clientId: String
     val clientSecret: String
 
-    def redirectLocation(req: Request): String = {
-      val hostStr = req.host.getOrElse(throw new Exception(s"Host not found in HTTP $req"))
+    def redirectLocation(req: Request, params: Tuple2[String, String]*): String = {
```
PeterFCaswell: Would it be safer to declare a type that is "param_name, param_value", rather than two random strings paired together? That way you could make sure that, for instance, the param_name is never empty.

auth/src/main/scala/com/lookout/borderpatrol/auth/tokenmaster/Tokenmaster.scala
```diff
@@ -74,21 +74,21 @@ object Tokenmaster {
 
     def aStringClaim(claim: String): String = wrapOps[String]({ () => accessClaimSet.getStringClaim(claim)},
       s"Failed to find '$claim' in the Access Token in the Request",
```
PeterFCaswell: Would it make sense for this error to read "Failed to find string claim '$claim' in the Access Token in the request"? And to put the type of the claim in the other errors too?

PeterFCaswell:
Those are the only comments I have at this point.

auth/src/main/scala/com/lookout/borderpatrol/auth/tokenmaster/Tokenmaster.scala
maheshkelkar: will do.

auth/src/main/scala/com/lookout/borderpatrol/auth/tokenmaster/LoginManagers.scala
maheshkelkar: this is common way of passing params as a tuple in twitter world.

Issue #44, opened 2015-08-02
Move from argonaut to jfc/circe


trane:
Argonaut is awesome, but it requires scalaz and more boilerplate.
We can seriously improve the code base and on-boarding by utilizing this port of Argonaut (argonaut-shapeless) to cats:

https://github.com/travisbrown/jfc

PR #160, opened 2016-06-08
Updated finagle version to 6.35.0, circe to 0.4.0, etc

### LINKED ISSUE #159: Update finagle version to 6.35.0
Finagle 6.35.0
TwitterServer 1.20.0

maheshkelkar:
Also, removed arganut and finch dependencies

Fixes #159, #44

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,13 +1,13 @@
 import sbtunidoc.Plugin.UnidocKeys._
 import scoverage.ScoverageSbtPlugin.ScoverageKeys.coverageExcludedPackages
 
-lazy val Version = "0.2.10-SNAPSHOT"
+lazy val Version = "0.2.11-SNAPSHOT"
 
 lazy val buildSettings = Seq(
   organization := "com.lookout",
   version := Version,
-  scalaVersion := "2.11.7",
-  crossScalaVersions := Seq("2.10.5", "2.11.7")
+  scalaVersion := "2.11.8",
+  crossScalaVersions := Seq("2.10.5", "2.11.8")
 )
 
 lazy val compilerOptions = Seq(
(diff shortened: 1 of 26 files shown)


build.sbt
```diff
@@ -56,7 +54,8 @@ val baseSettings = Seq(
       }
   ),
   scalacOptions in (Compile, console) := compilerOptions :+ "-Yrepl-class-based",
-  wartremoverWarnings in (Compile, compile) ++= Warts.allBut(Wart.FinalCaseClass, Wart.NoNeedForMonad, Wart.Throw, Wart.Null, Wart.Nothing, Wart.DefaultArguments)
+  wartremoverWarnings in (Compile, compile) ++= Warts.allBut(Wart.FinalCaseClass, Wart.NoNeedForMonad, Wart.Throw,
```
PeterFCaswell: Are these warnings disabled because our code generates them or some code that we pull in? Do you intend to get rid of them? Maybe in a future commit you could comment on why each is turned off?

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
```diff
@@ -275,7 +266,7 @@ object ConsulResponse {
   def apply(createIndex: Int, flags: Int, key: String, lockIndex: Int, modifyIndex: Int,
             secrets: Secrets): ConsulResponse =
     ConsulResponse(createIndex, flags, key, lockIndex, modifyIndex,
-      DatatypeConverter.printBase64Binary(SecretsEncoder.EncodeJson.encode(secrets).nospaces.getBytes))
+      DatatypeConverter.printBase64Binary(secrets.asJson.noSpaces.getBytes))
```
PeterFCaswell: Do we use noSpaces simply to save on the number of bytes? Or is there another reason?

build.sbt
maheshkelkar: This change was limited to fitting the line in 120 chars. I will look at this closely and let you know. This is something has been there since the beginging.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SecretStores.scala
maheshkelkar: noSpaces is just another variation of toString in `circe`. Makes it easier to copy and test the same command from command line

PR #167, opened 2016-07-05
On user error, redirect with error code instead of error message as query param

maheshkelkar:
When user login fails due to user error, then user is redirected back to login page. Currently, the redirect location contains query param to indicate failure message. Intent of this fix is to change this error message to a http status code

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,7 @@
 import sbtunidoc.Plugin.UnidocKeys._
 import scoverage.ScoverageSbtPlugin.ScoverageKeys.coverageExcludedPackages
 
-lazy val Version = "0.2.16-SNAPSHOT"
+lazy val Version = "0.2.17-SNAPSHOT"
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -1,7 +1,7 @@
 import sbtunidoc.Plugin.UnidocKeys._
 import scoverage.ScoverageSbtPlugin.ScoverageKeys.coverageExcludedPackages
 
-lazy val Version = "0.2.16-SNAPSHOT"
+lazy val Version = "0.2.17-SNAPSHOT"
```
PeterFCaswell: Isn't it time to delete the SNAPSHOT here?

maheshkelkar: In opensource, need it. Because SNAPSHOT versions goto jfrog (unstable) and stable versions goto bintray. I will cut a 0.3.0 version really soon

PR #170, opened 2016-07-13
Re-route forbidden requests to Borderpatrol

malikvivek:
Forbidden requests to Borderpatrol to be routed to forbidden page instead of login page.


core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -500,6 +500,7 @@ case class ExceptionFilter() extends SimpleFilter[Request, Response] {
     case error: BpUserError => logAndResponse(req, error.getMessage, error.status, Level.INFO)
     case error: BpCommunicationError => logAndResponse(req, error.getMessage, error.status, Level.WARNING)
     case error: BpCoreError => logAndResponse(req, error.getMessage, error.status, Level.INFO)
+    case error: BpForbiddenRequest => logAndResponse(req, error.getMessage, Status.Forbidden, Level.INFO)
```
maheshkelkar: Could you please add a scalatest for this change for `BpForbiddenRequest` in BorderAuthSpec.scala

PR #173, opened 2016-07-19
Add support for validating host.

rikesh-chouhan:
Added HostHeaderFilter to verify that the request being serviced has either empty
or valid (entry is contained in the filters valid hosts set) host entry

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,8 @@ in the HTTP response.
  * Secret Store:
    * A store is used to cache secret used to sign the session id(s)
  * Host Checker:
-   * A filter that blocks any request with a host entry, **if present** that is not present in the `allowedDomains` entry.
+   * A filter that blocks any request with a host entry that is not present in the `allowedDomains` entry.
+If request does not contain host header, this filter is a noop.
 This filter is applied by default and the config will throw an exception if there is *no entry/0* entries for this field.
 The code should be updated if you don't intend on using it.
 
@@ -159,8 +160,8 @@ this service
    * `prefix`: Prefix attached to each reported stat
  * `listeningPort`: Border Patrol listens to new requests on this port.
  * `healthCheckEndpoints`: A set of endpoints that impact the Border Patrol Health Status
- * `allowedDomains`: Border Patrol checks whether incoming request has a host header value **If present** 
-that is present in this set.
+ * `allowedDomains`: Border Patrol checks whether incoming request has a host header value present
+in this set.
 
      ```json
      "allowedDomains" : [ "my.localhost", "localhost.com", "yet.another.domain.com"]
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ Border Patrol Concepts
  * The identity endpoint, access endpoint, subdomains and default service forms the policy, which is defined using
 `LoginManager`.
  * When a HTTP request with URL `<http[s]>://<subdomain>:[port]/<path-prefix>` hits BorderPatrol:
+   * It can check whether or not a host specified is supported by this instance of BorderPatrol.
    * It looks up the CustomerIdentifier using subdomain prefix. It returns 404, if no match is found.
    * It looks up the ServiceIdentifier using path prefix. In the absence of any path (i.e. Root), the request is
 redirected to the default service.
@@ -154,6 +155,12 @@ this service
    * `prefix`: Prefix attached to each reported stat
  * `listeningPort`: Border Patrol listens to new requests on this port.
  * `healthCheckEndpoints`: A set of endpoints that impact the Border Patrol Health Status
+ * `allowedDomains`: Border Patrol checks whether incoming request has a host header value (*IF* present) 
+that is present in this set.
+
+     ```json
+     "allowedDomains" : [ "my.localhost", "localhost.com", "yet.another.domain.com"]
+     ```
 
 Modules
 -------
--- a/README.md
+++ b/README.md
@@ -68,6 +68,10 @@ in the HTTP response.
    * For authenticate user, the Session contains Master Token and one or more Service Token(s)
  * Secret Store:
    * A store is used to cache secret used to sign the session id(s)
+ * Host Checker:
+   * A filter that blocks any request with a host entry, **if present** that is not present in the `allowedDomains` entry.
+This filter is applied by default and the config will throw an exception if there is *no entry/0* entries for this field.
+The code should be updated if you don't intend on using it.
 
 Configuration
 -------------
@@ -155,7 +159,7 @@ this service
    * `prefix`: Prefix attached to each reported stat
  * `listeningPort`: Border Patrol listens to new requests on this port.
  * `healthCheckEndpoints`: A set of endpoints that impact the Border Patrol Health Status
- * `allowedDomains`: Border Patrol checks whether incoming request has a host header value (*IF* present) 
+ * `allowedDomains`: Border Patrol checks whether incoming request has a host header value **If present** 
 that is present in this set.
 
      ```json
--- a/bpConfig.json
+++ b/bpConfig.json
@@ -5,6 +5,7 @@
      "durationInSec": 60,
      "prefix": "bp"
   },
+  "whitelistedDomains": ["localhost", "my.localhost"],
   "sessionStore": {
     "type": "InMemoryStore"
   },
--- a/bpConfig.json
+++ b/bpConfig.json
@@ -5,7 +5,7 @@
      "durationInSec": 60,
      "prefix": "bp"
   },
-  "whitelistedDomains": ["localhost", "my.localhost"],
+  "allowedDomains": ["localhost", "my.localhost"],
   "sessionStore": {
     "type": "InMemoryStore"
   },
--- a/example/src/main/scala/com/lookout/borderpatrol/example/ServerConfig.scala
+++ b/example/src/main/scala/com/lookout/borderpatrol/example/ServerConfig.scala
@@ -16,7 +16,6 @@ import io.circe.generic.auto._
 import io.circe.syntax._
 import scala.io.Source
 
-
 case class StatsdExporterConfig(host: String, durationInSec: Int, prefix: String)
 
 
@@ -34,6 +33,7 @@ object EndpointConfig {
  * Server Config
  */
 case class ServerConfig(listeningPort: Int,
+                        whitelistedDomains: List[String],
                         secretStore: SecretStoreApi,
                         sessionStore: SessionStore,
                         statsdExporterConfig: StatsdExporterConfig,
@@ -166,6 +166,7 @@ object ServerConfig {
   implicit val serverConfigDecoder: Decoder[ServerConfig] = Decoder.instance { c =>
     for {
       listeningPort <- c.downField("listeningPort").as[Int]
+      whitelistedDomains <- c.downField("whiteListedDomains").as[List[String]]
       secretStore <- c.downField("secretStore").as[SecretStoreApi]
       sessionStore <- c.downField("sessionStore").as[SessionStore]
       statsdExporterConfig <- c.downField("statsdReporter").as[StatsdExporterConfig]
@@ -184,7 +185,7 @@ object ServerConfig {
           if (healthCheckEndpointOpts.contains(None)) None else Some(healthCheckEndpointOpts.flatten)
         },
         DecodingFailure(s"Failed to decode endpoint(s) in the healthCheckEndpoints: ", c.history))
-    } yield ServerConfig(listeningPort, secretStore, sessionStore, statsdExporterConfig,
+    } yield ServerConfig(listeningPort, whitelistedDomains, secretStore, sessionStore, statsdExporterConfig,
       healthCheckEndpointConfigs, cids, sids, lms, eps)
   }
 
@@ -226,7 +227,10 @@ object ServerConfig {
       validateServiceIdentifierConfig("serviceIdentifiers", serverConfig.serviceIdentifiers) ++
 
       //  Validate customerIdentifiers config
-      validateCustomerIdentifierConfig("customerIdentifiers", serverConfig.customerIdentifiers))
+      validateCustomerIdentifierConfig("customerIdentifiers", serverConfig.customerIdentifiers) ++
+
+      // Validate domain names config
+      validateDomainNames("whitelistedDomains", serverConfig.whitelistedDomains))
   }
 
   /**
--- a/example/src/main/scala/com/lookout/borderpatrol/example/ServerConfig.scala
+++ b/example/src/main/scala/com/lookout/borderpatrol/example/ServerConfig.scala
@@ -208,6 +208,16 @@ object ServerConfig {
   }
 
   /**
+    * Validate allowedDomains set to ensure that it is not empty
+    *
+    * @param field
+    * @param domainsSet
+    * @return error that the set is empty if domainsSet is empty
+    */
+  def validateAllowedDomains(field: String, domainsSet: Set[InternetDomainName]): Set[String] = {
+    cond(domainsSet.isEmpty, s"Cannot have empty set for field: '${field}'")
+  }
+  /**
    * Validates the BorderPatrol Configuration
    * - for duplicates
    * - invalid host configurations
@@ -229,7 +239,10 @@ object ServerConfig {
       validateServiceIdentifierConfig("serviceIdentifiers", serverConfig.serviceIdentifiers) ++
 
       //  Validate customerIdentifiers config
-      validateCustomerIdentifierConfig("customerIdentifiers", serverConfig.customerIdentifiers))
+      validateCustomerIdentifierConfig("customerIdentifiers", serverConfig.customerIdentifiers) ++
+
+      // Validate allowedDomains set
+      validateAllowedDomains("allowedDomains", serverConfig.allowedDomains))
   }
 
   /**
--- a/example/src/main/scala/com/lookout/borderpatrol/example/ServerConfig.scala
+++ b/example/src/main/scala/com/lookout/borderpatrol/example/ServerConfig.scala
@@ -2,6 +2,7 @@ package com.lookout.borderpatrol.example
 
 import java.net.URL
 
+import com.google.common.net.InternetDomainName
 import com.lookout.borderpatrol._
 import com.lookout.borderpatrol.auth.tokenmaster.LoginManagers.{BasicLoginManager, OAuth2LoginManager}
 import com.lookout.borderpatrol.server._
@@ -33,7 +34,7 @@ object EndpointConfig {
  * Server Config
  */
 case class ServerConfig(listeningPort: Int,
-                        whitelistedDomains: List[String],
+                        allowedDomains: Set[InternetDomainName],
                         secretStore: SecretStoreApi,
                         sessionStore: SessionStore,
                         statsdExporterConfig: StatsdExporterConfig,
@@ -166,7 +167,7 @@ object ServerConfig {
   implicit val serverConfigDecoder: Decoder[ServerConfig] = Decoder.instance { c =>
     for {
       listeningPort <- c.downField("listeningPort").as[Int]
-      whitelistedDomains <- c.downField("whiteListedDomains").as[List[String]]
+      allowedDomains <- c.downField("whiteListedDomains").as[Set[InternetDomainName]]
       secretStore <- c.downField("secretStore").as[SecretStoreApi]
       sessionStore <- c.downField("sessionStore").as[SessionStore]
       statsdExporterConfig <- c.downField("statsdReporter").as[StatsdExporterConfig]
@@ -185,7 +186,7 @@ object ServerConfig {
           if (healthCheckEndpointOpts.contains(None)) None else Some(healthCheckEndpointOpts.flatten)
         },
         DecodingFailure(s"Failed to decode endpoint(s) in the healthCheckEndpoints: ", c.history))
-    } yield ServerConfig(listeningPort, whitelistedDomains, secretStore, sessionStore, statsdExporterConfig,
+    } yield ServerConfig(listeningPort, allowedDomains, secretStore, sessionStore, statsdExporterConfig,
       healthCheckEndpointConfigs, cids, sids, lms, eps)
   }
 
@@ -227,10 +228,7 @@ object ServerConfig {
       validateServiceIdentifierConfig("serviceIdentifiers", serverConfig.serviceIdentifiers) ++
 
       //  Validate customerIdentifiers config
-      validateCustomerIdentifierConfig("customerIdentifiers", serverConfig.customerIdentifiers) ++
-
-      // Validate domain names config
-      validateDomainNames("whitelistedDomains", serverConfig.whitelistedDomains))
+      validateCustomerIdentifierConfig("customerIdentifiers", serverConfig.customerIdentifiers))
   }
 
   /**
--- a/example/src/main/scala/com/lookout/borderpatrol/example/ServerConfig.scala
+++ b/example/src/main/scala/com/lookout/borderpatrol/example/ServerConfig.scala
@@ -154,6 +154,7 @@ object ServerConfig {
   implicit val serverConfigEncoder: Encoder[ServerConfig] = Encoder.instance { serverConfig =>
     Json.fromFields(Seq(
       ("listeningPort", serverConfig.listeningPort.asJson),
+      ("allowedDomains", serverConfig.allowedDomains.asJson),
       ("secretStore", serverConfig.secretStore.asJson),
       ("sessionStore", serverConfig.sessionStore.asJson),
       ("statsdReporter", serverConfig.statsdExporterConfig.asJson),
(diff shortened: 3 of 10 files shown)


bpConfig.json
```diff
@@ -5,6 +5,7 @@
      "durationInSec": 60,
      "prefix": "bp"
   },
+  "whitelistedDomains": ["localhost", "my.localhost"],
```
PeterFCaswell: Any chance that we can use a term other than "whitelist"?  For instance, instead of "blackList", we can use "blockList" and use "clearList" or similar for "whitelist". Black == bad & white == good is probably something we shouldn't perpetuate.

PeterFCaswell:
I'm fine with this other than the name of the list. :-)  Sorry to be a pain, but it is a real thing.

example/src/main/scala/com/lookout/borderpatrol/example/ServerConfig.scala
```diff
@@ -34,6 +33,7 @@ object EndpointConfig {
  * Server Config
  */
 case class ServerConfig(listeningPort: Int,
+                        whitelistedDomains: List[String],
```
maheshkelkar: can we use `Set` to catch duplicates

example/src/main/scala/com/lookout/borderpatrol/example/ServerConfig.scala
```diff
@@ -166,6 +166,7 @@ object ServerConfig {
   implicit val serverConfigDecoder: Decoder[ServerConfig] = Decoder.instance { c =>
     for {
       listeningPort <- c.downField("listeningPort").as[Int]
+      whitelistedDomains <- c.downField("whiteListedDomains").as[List[String]]
```
maheshkelkar: `Set`

server/src/main/scala/com/lookout/borderpatrol/server/Config.scala
```diff
@@ -193,4 +194,22 @@ object Config {
...
+  /**
+    * Validate domain name(s) specified for some field
+    *
+    * @param field - used for error reporting
+    * @param domainNames - set of names that are to be used as domains
+    * @return set of all errors encountered during validation
+    */
+  def validateDomainNames(field: String, domainNames: List[String]): Set[String] = {
+    // Find if endpoints have duplicate entries
+    ((cond(domainNames.distinct.size != domainNames.size,
```
maheshkelkar: we dont need it, if we use `Set`

server/src/main/scala/com/lookout/borderpatrol/server/Config.scala
```diff
@@ -12,6 +12,7 @@ import cats.data.Xor
 import com.twitter.logging.Logger
 import io.circe.{Encoder, _}
 import io.circe.syntax._
+import com.google.common.net.InternetDomainName
```
maheshkelkar: does this require us to add a dependency? are we pulling in a world for that?

server/src/main/scala/com/lookout/borderpatrol/server/Config.scala
```diff
@@ -193,4 +194,22 @@ object Config {
...
+    * @param field - used for error reporting
+    * @param domainNames - set of names that are to be used as domains
+    * @return set of all errors encountered during validation
+    */
+  def validateDomainNames(field: String, domainNames: List[String]): Set[String] = {
+    // Find if endpoints have duplicate entries
+    ((cond(domainNames.distinct.size != domainNames.size,
+      s"Duplicate entries for key (domain name) are found in the field: ${field}")) ++
+
+    domainNames.filter(domainName => !InternetDomainName.isValid(domainName)) collect { case k =>
```
maheshkelkar: Instead of `List[String]`, why not use `Set[InternalDomainName]`?

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
```diff
@@ -0,0 +1,42 @@
...
+  /**
+    * Check that the request received contains a host entry [if present] that is present in the
+    * validHosts set of names specified to this filter. If not then return an empty Future[Response]
+    * or error
+    *
+    * @param validHosts
+    */
+  case class HostChecker(validHosts: Set[String]) extends SimpleFilter[Request, Response] {
+
+    private def checkHostEntry(request: Request): Request =
```
maheshkelkar: `private[this]`

server/src/main/scala/com/lookout/borderpatrol/server/Config.scala
rikesh-chouhan: no. guava library is already part of the dependencies [there are no changes to sbt file as part of this commit]

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
```diff
@@ -0,0 +1,42 @@
...
+      tap(request) { re =>
+        re.host match {
+          case Some(s) if !s.isEmpty => if (!validHosts(s))
+            throw new BpUserError(Status.NotFound, s"Host Header: $s not found")
```
maheshkelkar: Please add single quote around the $s
i.e. `'$s'`

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
```diff
@@ -0,0 +1,42 @@
...
+    * or error
+    *
+    * @param validHosts
+    */
+  case class HostChecker(validHosts: Set[String]) extends SimpleFilter[Request, Response] {
+
+    private def checkHostEntry(request: Request): Request =
+      tap(request) { re =>
+        re.host match {
+          case Some(s) if !s.isEmpty => if (!validHosts(s))
```
maheshkelkar: is `s.isEmpty` check necessary? Can `validHosts(s)` perform efficiently, if `s` is empty?

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
```diff
@@ -0,0 +1,42 @@
...
+    * Check that the request received contains a host entry [if present] that is present in the
+    * validHosts set of names specified to this filter. If not then return an empty Future[Response]
+    * or error
+    *
+    * @param validHosts
+    */
+  case class HostChecker(validHosts: Set[String]) extends SimpleFilter[Request, Response] {
+
+    private def checkHostEntry(request: Request): Request =
+      tap(request) { re =>
```
maheshkelkar: we do not need `tap` because we are not modifying the request. We are simply checking if contents of request are valid. Take a look at SecureHeadersFilter for example.

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
```diff
@@ -0,0 +1,42 @@
...
+        re.host match {
+          case Some(s) if !s.isEmpty => if (!validHosts(s))
+            throw new BpUserError(Status.NotFound, s"Host Header: $s not found")
+          case _ => (re)
+        }
+    }
+
+
+    /**
+      * Requests get X-Forwarded-For and other request headers added before passing to the service
```
maheshkelkar: stale comments

security/src/main/scala/com/lookout/borderpatrol/security/SecureHeaders.scala
```diff
@@ -19,7 +19,7 @@ object SecureHeaders {
   val XXSSProtection = ("X-XSS-Protection", "1; mode=block")
   val XContentTypeOptions = ("X-ContentType-Options", "nosniff")
   val XDownloadOptions = ("X-Download-Options", "noopen")
-  val XPermittedCrossDomainPolicies = ("X-Permitted-Cross-Domain-Policies", "none")
+  var XPermittedCrossDomainPolicies = ("X-Permitted-Cross-Domain-Policies", "none")
```
maheshkelkar: we dont want to use `var` here

maheshkelkar:
Could you please update `build.sbt` to bump up the version? BTW if you commit, before me, then please skip `0.2.22-SNAPSHOT` version.

bpConfig.json
rikesh-chouhan: done - updating to allowedDomains

server/src/main/scala/com/lookout/borderpatrol/server/Config.scala
rikesh-chouhan: Changed to Set[String].

server/src/test/scala/com/lookout/borderpatrol/server/ConfigSpec.scala
```diff
@@ -187,9 +188,12 @@ class ConfigSpec extends BorderPatrolSuite {
       "hosts configuration for failed4 in some: has unsupported protocol")
   }
 
-  it should "validate domain names" in {
-    val domainNames = List("www.bluelotussoftware.com", "bluelotussoftware.com", "bluelotussoftware", "256.0.0.0")
+  it should "uphold encode decode InternetDomainName to string" in {
+    val lookout = InternetDomainName.from("source.corp.lookout.com")
```
maheshkelkar: no lookout here.

server/src/test/scala/com/lookout/borderpatrol/server/ConfigSpec.scala
```diff
@@ -187,9 +188,12 @@ class ConfigSpec extends BorderPatrolSuite {
...
+    decodeInternetDomainName.decodeJson(lookout.asJson).toOption.get should be (lookout)
 
-    validateDomainNames("whitelistDomains", domainNames).toString().contains("256") should be (true)
+    val yahoo = InternetDomainName.from("yahoo.com")
+    encodeInternetDomainName(yahoo) eq("yahoo.com")
```
maheshkelkar: look at the other examples, we should calls `decode(encode(domain)) should be(domain)`

maheshkelkar:
@rikesh-chouhan: I think there are still few comments in filter outstanding

maheshkelkar:
@rikesh-chouhan: it looks like there are few `scalastyle` errors

example/src/main/scala/com/lookout/borderpatrol/example/ServerConfig.scala
rikesh-chouhan: Done

example/src/main/scala/com/lookout/borderpatrol/example/ServerConfig.scala
rikesh-chouhan: Done

server/src/main/scala/com/lookout/borderpatrol/server/Config.scala
rikesh-chouhan: Done

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
rikesh-chouhan: updated.

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
rikesh-chouhan: SecureHeaderFilter uses tap. If I don't use tap - I get an error in pattern matching for "return type cannot be Unit"

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
rikesh-chouhan: done

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
rikesh-chouhan: done.

server/src/test/scala/com/lookout/borderpatrol/server/ConfigSpec.scala
```diff
@@ -186,4 +187,13 @@ class ConfigSpec extends BorderPatrolSuite {
     validateHostsConfig("some", "failed4", Set()) should contain (
       "hosts configuration for failed4 in some: has unsupported protocol")
   }
+
+  it should "uphold encode decode InternetDomainName to string" in {
+    val lookout = InternetDomainName.from("source.corp.wiki.com")
+    decodeInternetDomainName.decodeJson(lookout.asJson).toOption.get should be (lookout)
+
+    val yahoo = InternetDomainName.from("yahoo.com")
+    encodeInternetDomainName(yahoo) eq("yahoo.com")
```
maheshkelkar: not sure what we are testing here. I don't think `eq` is working here.

Also, can you please add failure cases to make sure its throwing exceptions

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
```diff
@@ -0,0 +1,40 @@
...
+    * validHosts set of names specified to this filter. If not then return an empty Future[Response]
+    * or error
+    *
+    * @param validHosts
+    */
+  case class HostChecker(validHosts: Set[String]) extends SimpleFilter[Request, Response] {
+
+    private[this] def checkHostEntry(request: Request): Request = {
+      val host = request.host
+      if (host.isDefined) {
```
maheshkelkar: instead of if-else, we should make it more functional by using req.host.foreach or match

```
req host match {
  Some(h) if !validHosts(h) => throw exception
  None => 
}
```

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
```diff
@@ -0,0 +1,40 @@
...
+  /**
+    * Check that the request received contains a host entry [if present] that is present in the
+    * validHosts set of names specified to this filter. If not then return an empty Future[Response]
+    * or error
+    *
+    * @param validHosts
+    */
+  case class HostChecker(validHosts: Set[String]) extends SimpleFilter[Request, Response] {
+
+    private[this] def checkHostEntry(request: Request): Request = {
```
maheshkelkar: The return type should be `Unit`

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
```diff
@@ -0,0 +1,40 @@
...
+import com.lookout.borderpatrol.auth.{BpUserError}
+import com.lookout.borderpatrol.util.Combinators._
+import com.twitter.finagle.{Service, SimpleFilter}
+import com.twitter.finagle.http.{Response, Request, Status}
+import com.twitter.util.Future
+
+/**
+  * Created by rikesh.chouhan on 7/18/16.
+  */
+object HostHeaderFilter {
```
maheshkelkar: Typically we should use `Filter` in the name of the Filter. So I would rather remove this object. And rename the `case class` and move it out of any object

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
```diff
@@ -0,0 +1,40 @@
...
+    * or error
+    *
+    * @param validHosts
+    */
+  case class HostChecker(validHosts: Set[String]) extends SimpleFilter[Request, Response] {
+
+    private[this] def checkHostEntry(request: Request): Request = {
+      val host = request.host
+      if (host.isDefined) {
+        if (!validHosts(host.get))
```
maheshkelkar: Is it fair to say that `validHosts(string)` will do the exact match on the string?

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
```diff
@@ -0,0 +1,40 @@
...
+        throw new BpUserError(Status.NotFound, s"Host Header: '${host.get}' not found")
```
maheshkelkar: We are using BpUserError for the recoverable errors that will redirect the user back to login page. Whereas this is unrecoverable, hence we should add a new error code. Add it to BpCoreError or BpAuthError and put error code in it. Similarly update BorderAuth.scala to take action on it

example/src/main/scala/com/lookout/borderpatrol/example/ServerConfig.scala
```diff
@@ -184,7 +187,7 @@ object ServerConfig {
           if (healthCheckEndpointOpts.contains(None)) None else Some(healthCheckEndpointOpts.flatten)
         },
         DecodingFailure(s"Failed to decode endpoint(s) in the healthCheckEndpoints: ", c.history))
-    } yield ServerConfig(listeningPort, secretStore, sessionStore, statsdExporterConfig,
+    } yield ServerConfig(listeningPort, allowedDomains, secretStore, sessionStore, statsdExporterConfig,
```
maheshkelkar: Could you please update Readme.md with this new field?

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
```diff
@@ -0,0 +1,40 @@
...
+object HostHeaderFilter {
+
+  /**
+    * Check that the request received contains a host entry [if present] that is present in the
+    * validHosts set of names specified to this filter. If not then return an empty Future[Response]
+    * or error
+    *
+    * @param validHosts
+    */
+  case class HostChecker(validHosts: Set[String]) extends SimpleFilter[Request, Response] {
```
maheshkelkar: How about using `Set[InternetDomainName]` instead of `Set[String]` so you dont need double conversions. And you don't have to add a check to validate contents of Set[String]

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
rikesh-chouhan: yes.

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
rikesh-chouhan: I would leave this as String, since, request.host returns Option[String]. I don't think we should force a type on request.host value.

maheshkelkar: I think you can easily address that with:
`req.host.foreach(h => validHosts.contains(InternetDomainNames.from(h))`

If not, you should add testcase with vairous values in `validHosts` array. Empty, non-host values, etc

rikesh-chouhan: `validHosts`is assigned after it has been vetted. I added a check for `validHosts` to be not empty. 
I'll update `Set[String]` to `Set[InternetDomainName]`.
The updated code would look like so -
`req.host.foreach(providedHost => validHosts map { validHost => validHost.toString(providedHost) })
I'm concerned about imposing the type **InternetDomainName**  _host_. If you can point documentation in **Finagle** that it verifies _host_ value I will gladly update it.

README.md
```diff
@@ -24,6 +24,7 @@ Border Patrol Concepts
  * The identity endpoint, access endpoint, subdomains and default service forms the policy, which is defined using
 `LoginManager`.
  * When a HTTP request with URL `<http[s]>://<subdomain>:[port]/<path-prefix>` hits BorderPatrol:
```
maheshkelkar: could you please change the `subdomain` to `host`?

example/src/main/scala/com/lookout/borderpatrol/example/service.scala
```diff
@@ -77,7 +77,7 @@ object service {
                        secretStore: SecretStoreApi): Service[Request, Response] = {
     val serviceMatcher = ServiceMatcher(config.customerIdentifiers, config.serviceIdentifiers)
     val notFoundService = Service.mk[SessionIdRequest, Response] { req => Response(Status.NotFound).toFuture }
-    val validHosts = config.allowedDomains.map( k => k.toString)
+    val validHosts = config.allowedDomains
```
maheshkelkar: skip this. just use the `config.allowedDomains` in the HeaderFilter initialization

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
```diff
@@ -1,40 +1,38 @@
...
 
-  /**
-    * Check that the request received contains a host entry [if present] that is present in the
-    * validHosts set of names specified to this filter. If not then return an empty Future[Response]
-    * or error
-    *
-    * @param validHosts
-    */
-  case class HostChecker(validHosts: Set[String]) extends SimpleFilter[Request, Response] {
+  if (validHosts.size == 0) {
```
maheshkelkar: why? I know its wrong. But, our validate config should ensure that. And if the array is null, then HostHeaderFilter will block everything.

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
```diff
@@ -1,40 +1,38 @@
...
-    private[this] def checkHostEntry(request: Request): Request = {
-      if (!validHosts.isEmpty && request.host.isDefined) {
-        val host = request.host
-        if (!validHosts(host.get))
-          throw new BpUserError(Status.NotFound, s"Host Header: '${host.get}' not found")
-      }
-      request
-    }
+  private[this] def checkHostEntry(request: Request): Unit = {
+    request.host.foreach( host => if (!validHosts.map( validHost => validHost.toString ).contains(host))
```
maheshkelkar: how about adding a `lazy val validHostStrings = validHosts.map( validHost => validHost.toString )`

server/src/test/scala/com/lookout/borderpatrol/server/ConfigSpec.scala
```diff
@@ -189,11 +189,18 @@ class ConfigSpec extends BorderPatrolSuite {
...
     val yahoo = InternetDomainName.from("yahoo.com")
-    encodeInternetDomainName(yahoo) eq("yahoo.com")
+    encodeInternetDomainName(yahoo) should be eq ("yahoo.com")
+
+    a[IllegalArgumentException]should be thrownBy {
+      decodeInternetDomainName.decodeJson(500.toString.asJson)
+    }
   }
 
+  private def decodedInternetDomainNameJson(internetDomainName: InternetDomainName): Option[InternetDomainName] = {
```
maheshkelkar: you can code it use it as follows:
```
it should "uphold encode decode InternetDomainName to string" in {
  def decodedInternetDomainNameJson(internetDomainName: InternetDomainName): Option[InternetDomainName] =
    decodeInternetDomainName.decodeJson(internetDomainName.asJson).toOption
  ::
  ::
```

maheshkelkar:
A few minor changes, otherwise it looks good.

server/src/test/scala/com/lookout/borderpatrol/server/ConfigSpec.scala
rikesh-chouhan: done.

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
rikesh-chouhan: updated.

README.md
```diff
@@ -68,6 +68,10 @@ in the HTTP response.
    * For authenticate user, the Session contains Master Token and one or more Service Token(s)
  * Secret Store:
    * A store is used to cache secret used to sign the session id(s)
+ * Host Checker:
+   * A filter that blocks any request with a host entry, **if present** that is not present in the `allowedDomains` entry.
```
maheshkelkar: this line is confusing, can you reword?

Perhaps **if present** is not necessary. if whether `array is empty` or `request.host doesnt exists in array` means the same thing.

README.md
```diff
@@ -155,7 +159,7 @@ this service
    * `prefix`: Prefix attached to each reported stat
  * `listeningPort`: Border Patrol listens to new requests on this port.
  * `healthCheckEndpoints`: A set of endpoints that impact the Border Patrol Health Status
- * `allowedDomains`: Border Patrol checks whether incoming request has a host header value (*IF* present) 
+ * `allowedDomains`: Border Patrol checks whether incoming request has a host header value **If present**
```
maheshkelkar: same as above

server/src/test/scala/com/lookout/borderpatrol/server/ConfigSpec.scala
```diff
@@ -186,4 +187,19 @@ class ConfigSpec extends BorderPatrolSuite {
...
+
+  it should "uphold encode decode InternetDomainName to string" in {
+    def decodedInternetDomainNameJson(internetDomainName: InternetDomainName): Option[InternetDomainName] =
+      decodeInternetDomainName.decodeJson(internetDomainName.asJson).toOption
+
+    val wiki = InternetDomainName.from("source.corp.wiki.com")
+    decodedInternetDomainNameJson(wiki).get should be (wiki)
+
+    val yahoo = InternetDomainName.from("yahoo.com")
+    encodeInternetDomainName(yahoo) should be eq ("yahoo.com")
```
maheshkelkar: do you want to change this to use `decodedInternalDomainNameJson`?

rikesh-chouhan: which line are you referring to?

README.md
rikesh-chouhan: rewording.

README.md
rikesh-chouhan: rewording

PR #174, opened 2016-07-19
Update to fiangle 6.36.0 version

maheshkelkar:
- fix exception raised due to change from apache codec to MIME encoder/decoder

### DIFF
--- a/core/src/main/scala/com/lookout/borderpatrol/sessionx/SignedId.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/sessionx/SignedId.scala
@@ -173,23 +173,16 @@ object SignedId {
       case _ => Failure(new BpSignedIdError("Not a session string"))
     }
 
-    implicit def seq2SignedId(bytes: IndexedSeq[Byte])(implicit store: SecretStoreApi): Try[SignedId] = for {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/com/lookout/borderpatrol/sessionx/SignedId.scala
```diff
@@ -173,23 +173,16 @@ object SignedId {
       case _ => Failure(new BpSignedIdError("Not a session string"))
     }
 
-    implicit def seq2SignedId(bytes: IndexedSeq[Byte])(implicit store: SecretStoreApi): Try[SignedId] = for {
+    def str2seq(s: String): Try[IndexedSeq[Byte]] =
```
PeterFCaswell: Why do we use Try here?  Can you use an Either instead?

maheshkelkar: Try catches the exception, whereas IIRC Either simply provides options. And the for comprehension composing all the `Try`s, hence using it here

PeterFCaswell: Ah, I misunderstood what Try was doing. Thanks.

PR #176, opened 2016-07-25
Add port stripping to host name obtained from request

rikesh-chouhan:


bpConfig.json
```diff
@@ -5,7 +5,7 @@
      "durationInSec": 60,
      "prefix": "bp"
   },
-  "allowedDomains": ["localhost", "my.localhost"],
+  "allowedDomains": ["api.localhost", "ent.localhost", "localhost"],
```
maheshkelkar: We should skip `"localhost"`

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
```diff
@@ -19,12 +19,31 @@ case class HostHeaderFilter(validHosts: Set[InternetDomainName]) extends SimpleF
...
     )
   }
 
   /**
+    * Strip out the port portion including the semicolon from the provided
+    * host entry.
+    * @param host string to strip port from
+    * @return empty or stripped port string
+    */
+  def filterPort(host: String): String = {
```
maheshkelkar: lets change it to return `Option[String]`

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
```diff
@@ -19,12 +19,31 @@ case class HostHeaderFilter(validHosts: Set[InternetDomainName]) extends SimpleF
...
+    if (host.contains(":") && host.indexOf(":") != 0)
+      host.substring(0, host.indexOf(":"))
+    else if (!host.contains(":"))
+      host
+    else /* the entry contains a ':' at the beginning of the string */
```
maheshkelkar: we should use `match` instead of if-else conditions

could you please find functional equivalent of the following:

```
if (!host.contains(":"))
  Some(host)
else if (host.indexOf(":") != 0)
  Some(host.substring(0, host.indexOf(":")))
else 
  None
}
```
But, we should find a better code `functional` code for this

security/src/test/scala/com/lookout/borderpatrol/test/security/HostHeaderFilterSpec.scala
```diff
@@ -38,4 +38,12 @@ class HostHeaderFilterSpec extends BorderPatrolSuite {
     caught.status should be (Status.NotFound)
   }
 
+  it should "return filter out port entries when present" in {
+    val someHostNames = Seq("hello.com", "yahoo.com:8080", ":dummy", " ", "localhost:8080")
+    val filteredNames: Seq[String] = someHostNames.map( s => checker.filterPort(s))
+    filteredNames.contains(":dummy") should be (false)
+    filteredNames.contains("hello.com") should be (true)
+    filteredNames.contains("localhost") should be (true)
+    filteredNames.contains("yahoo.com") should be (true)
```
maheshkelkar: add 1 test with numbers

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
maheshkelkar: I think you kept it public for testing. You can either make it protected or private to this package by using private[package name]

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
```diff
@@ -19,12 +19,31 @@ case class HostHeaderFilter(validHosts: Set[InternetDomainName]) extends SimpleF
   lazy val validHostStrings = validHosts.map( validHost => validHost.toString )
 
   private[this] def checkHostEntry(request: Request): Unit = {
-    request.host.foreach( host => if (!validHostStrings.contains(host))
-      throw new BpNotFoundRequest(s"Host Header: '${host}' not found")
+    request.host.foreach( host => {
+      val hostNameOnly = filterPort(host).trim
+      if (hostNameOnly.length > 0 && !validHostStrings.contains(hostNameOnly)) {
+        throw new BpNotFoundRequest(s"Host Header: '${hostNameOnly}' not found")
```
maheshkelkar: BTW, why didn't you use parseHostPorts?

once you change filterPort to return option, it should become:

```
filterPort(host).trim match {
  case Some(h) if !validHostStrings.contains(hostNameOnly) => throw ...
  case _ => 
}
```

bpConfig.json
rikesh-chouhan: done

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
rikesh-chouhan: done

security/src/test/scala/com/lookout/borderpatrol/test/security/HostHeaderFilterSpec.scala
rikesh-chouhan: this test is changing - FYI.

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
rikesh-chouhan: changing it.

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
```diff
@@ -18,29 +20,28 @@ case class HostHeaderFilter(validHosts: Set[InternetDomainName]) extends SimpleF
...
     */
-  def filterPort(host: String): String = {
-    if (host.contains(":") && host.indexOf(":") != 0)
-      host.substring(0, host.indexOf(":"))
-    else if (!host.contains(":"))
-      host
-    else /* the entry contains a ':' at the beginning of the string */
-      ""
+  private[security] def extractHostName(host: String): Option[String] = {
+    Try (InetSocketAddressUtil.parseHostPorts(host).seq.head._1) match {
```
maheshkelkar: I am not sure if this will work. You may want to use `headOption` instead of `seq.head._1` and then you are matching `Try[Option[String]]`

maheshkelkar: Here is another alternative:
```
scala> def mIt(str: String): Option[String] = str.split(":") match {
     | case Array(host) if host.nonEmpty => Some(host)
     | case Array(host, port) if host.nonEmpty => Some(host)
     | case _ => None
     | }
mIt: (str: String)Option[String]

scala> mIt(":foo")
res7: Option[String] = None

scala> mIt(":")
res8: Option[String] = None

scala> mIt("a:")
res9: Option[String] = Some(a)

scala> mIt("a")
res10: Option[String] = Some(a)

scala> 

```

rikesh-chouhan: This alternativ3 do3s not use parseHostPorts.

rikesh-chouhan: > I am not sure if this will work. You may want to use headOption instead of seq.head._1 and then you are matching Try[Option[String]]

It does work. Which case does it fail?

PR #178, opened 2016-07-28
Fix mock services to handle content in json format

maheshkelkar:
Update the mock to process the json content. Also, updated the formatRedirectResponse code to take `msgSrc` as an optional input


rikesh-chouhan:
+1

PR #180, opened 2016-07-28
HostHeaderFilter checks for empty host and returns 400

rikesh-chouhan:
Merge successful

### DIFF
--- a/example/src/main/scala/com/lookout/borderpatrol/example/service.scala
+++ b/example/src/main/scala/com/lookout/borderpatrol/example/service.scala
@@ -78,10 +78,10 @@ object service {
     val serviceMatcher = ServiceMatcher(config.customerIdentifiers, config.serviceIdentifiers)
     val notFoundService = Service.mk[SessionIdRequest, Response] { req => Response(Status.NotFound).toFuture }
     val serviceChainFront: Filter[Request, Response, Request, Response] =
+    /* Validate host if present to be present in pre-configured list*/
+      HostHeaderFilter(config.allowedDomains) andThen
       /* Convert exceptions to responses */
-      ExceptionFilter() andThen
-        /* Validate host if present to be present in pre-configured list*/
-        HostHeaderFilter(config.allowedDomains)
+      ExceptionFilter()
 
     RoutingService.byPath {
       case "/health" =>
--- a/security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
+++ b/security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
@@ -1,12 +1,13 @@
 package com.lookout.borderpatrol.security
 
 import com.google.common.net.InternetDomainName
-import com.lookout.borderpatrol.BpNotFoundRequest
+import com.lookout.borderpatrol.util.Combinators.tap
 import com.twitter.finagle.util.InetSocketAddressUtil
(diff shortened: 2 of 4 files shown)


security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
```diff
@@ -31,21 +33,32 @@ case class HostHeaderFilter(validHosts: Set[InternetDomainName]) extends SimpleF
...
+        log.log(Level.WARNING, "Host Header: is empty")
+        cannedResponse
+      case Some(hostName) if (!validHostStrings(hostName)) =>
+        log.log(Level.WARNING, s"Host Header: '${hostName}' not found")
+        cannedResponse
+      case _ => service(request)
+    }
+  }
+
+  private[this] def cannedResponse(): Future[Response] = {
```
maheshkelkar: this should be `val` instead of method

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
```diff
@@ -31,21 +33,32 @@ case class HostHeaderFilter(validHosts: Set[InternetDomainName]) extends SimpleF
...
     */
-  def apply(req: Request, service: Service[Request, Response]): Future[Response] = {
-    checkHostEntry(req)
-    for {
-      resp <- service(req)
-    } yield resp
+  def apply(request: Request, service: Service[Request, Response]): Future[Response] = {
+    request.host.map( extractHostName(_)) match {
+      case None =>
+        log.log(Level.WARNING, "Host Header: is empty")
```
maheshkelkar: I don't think this log adds any value here.

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
```diff
@@ -31,21 +33,32 @@ case class HostHeaderFilter(validHosts: Set[InternetDomainName]) extends SimpleF
...
+        cannedResponse
+      case Some(hostName) if (!validHostStrings(hostName)) =>
+        log.log(Level.WARNING, s"Host Header: '${hostName}' not found")
```
maheshkelkar: same as above

security/src/main/scala/com/lookout/borderpatrol/security/HostHeaderFilter.scala
rikesh-chouhan: I can remove it. My thought is, if this is debug, we could count how many times we get host header as empty. 
We could use datadog or something for count though.

example/src/main/scala/com/lookout/borderpatrol/example/service.scala
```diff
@@ -78,10 +78,10 @@ object service {
     val serviceMatcher = ServiceMatcher(config.customerIdentifiers, config.serviceIdentifiers)
     val notFoundService = Service.mk[SessionIdRequest, Response] { req => Response(Status.NotFound).toFuture }
     val serviceChainFront: Filter[Request, Response, Request, Response] =
+    /* Validate host if present to be present in pre-configured list*/
```
PeterFCaswell: When you modify this again, can you indent this comment two bytes, please?

rikesh-chouhan: sure. I've to make some updates.

PR #181, opened 2016-07-28
Treat BpOriginalRequestNotFound as a BpUserEror

maheshkelkar:
- redirect user to login page on this error
- added logs on failure to lookup or decode session data

### DIFF
--- a/core/src/main/scala/com/lookout/borderpatrol/sessionx/SessionStore.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/sessionx/SessionStore.scala
@@ -1,13 +1,14 @@
 package com.lookout.borderpatrol.sessionx
 
-import com.lookout.borderpatrol.{HealthCheckStatus, HealthCheck}
+import com.lookout.borderpatrol.{HealthCheck, HealthCheckStatus}
 import com.twitter.finagle.http.Status
 import com.twitter.io.Buf
-import com.twitter.util.{Time, Future}
+import com.twitter.util.{Future, Time}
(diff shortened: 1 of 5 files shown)


core/src/main/scala/com/lookout/borderpatrol/sessionx/SessionStore.scala
```diff
@@ -47,17 +50,21 @@ object SessionStores {
...
     def get[A](key: SignedId)(implicit ev: SessionDataEncoder[A]): Future[Option[Session[A]]] =
-      store.get(key.asBase64).flatMap(_ match {
-        case None => Future.value(None)
-        case Some(buf) => ev.decode(buf) match {
-          case Failure(e) => e.toFutureException[Option[Session[A]]]
-          case Success(data) => Some(Session(key, data)).toFuture
-        }
-      })
+      store.get(key.asBase64).flatMap {
+        case None =>
```
PeterFCaswell: Isn't it more normal to see the 'case None' as the last case in the match?

rikesh-chouhan: I think in this case it is an indicator for "empty" value. And it would be preferable to note what the behavior is in such a situation right off the bat.

PeterFCaswell: okay

rikesh-chouhan:
+1

PeterFCaswell:
+2

PR #183, opened 2016-08-01
Add support for pushing CORS headers in response

rikesh-chouhan:
Add CORS support. Update tests for specific expectations - such as overriding params in response and presence of CORS params in response.

### DIFF
--- a/security/src/main/scala/com/lookout/borderpatrol/security/SecureHeaders.scala
+++ b/security/src/main/scala/com/lookout/borderpatrol/security/SecureHeaders.scala
@@ -2,6 +2,7 @@ package com.lookout.borderpatrol.security
 
 import java.net.InetAddress
 
+import com.google.common.net.InternetDomainName
 import com.lookout.borderpatrol.sessionx._
 import com.lookout.borderpatrol.util.Combinators.tap
 import com.twitter.finagle.{Service, SimpleFilter}
@@ -13,6 +14,7 @@ import com.twitter.util.Future
   */
 object SecureHeaders {
 
+  val allowOrigin = "Access-Control-Allow-Origin"
   // response headers https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Response_fields
   val StrictTransportSecurity = ("Strict-Transport-Security", "max-age=31557600")
   val XFrameOptions = ("X-Frame-Options", "DENY")
@@ -20,9 +22,16 @@ object SecureHeaders {
   val XContentTypeOptions = ("X-ContentType-Options", "nosniff")
   val XDownloadOptions = ("X-Download-Options", "noopen")
   val XPermittedCrossDomainPolicies = ("X-Permitted-Cross-Domain-Policies", "none")
+  val controlEnabled = "Access-Control-Allow-Credentials" -> "true"
+  val exposeHeaders = "Access-Control-Expose-Headers" -> "ETag,Set-Cookie"
+  val allowMethods = "Access-Control-Allow-Methods" -> "GET, POST, PATCH, PUT, DELETE, OPTIONS, HEAD"
+  val allowHeaders =
+    "Access-Control-Allow-Headers" -> "*,x-csrf-token,x-requested-with,Content-Type,If-Modified-Since,If-None-Match"
+  val maxAge = "Access-Control-Max-Age" -> "86400"
 
   val response = HeaderMap(StrictTransportSecurity, XFrameOptions, XXSSProtection,
-                           XContentTypeOptions, XDownloadOptions, XPermittedCrossDomainPolicies)
+                           XContentTypeOptions, XDownloadOptions, XPermittedCrossDomainPolicies,
+                           allowMethods, allowHeaders, controlEnabled,exposeHeaders, maxAge)
   val request = HeaderMap()
 }
 
@@ -49,15 +58,24 @@ object SecureHeaders {
   * @param responseHeaders
   */
 case class SecureHeaderFilter(requestHeaders: HeaderMap = SecureHeaders.request,
-                              responseHeaders: HeaderMap = SecureHeaders.response)
+                              responseHeaders: HeaderMap = SecureHeaders.response,
+                              allowedDomains: Set[InternetDomainName])
     extends SimpleFilter[Request, Response] {
+  val validHostsString = allowedDomains.mkString(",")
   val localIp = InetAddress.getLocalHost.getHostAddress
+  val controlOrigin = SecureHeaders.allowOrigin -> validHostsString
 
   def injectRequestHeaders(req: Request): Request =
     tap(req) { re =>
       re.headerMap ++= requestHeaders
       re.xForwardedFor_=(re.xForwardedFor.fold(localIp)(s => s"$s, $localIp"))
     }
+
+  private[this] def injectResponseHeaders(response: Response): Response = {
+    response.headerMap ++= responseHeaders
+    response.headerMap.add(controlOrigin._1, controlOrigin._2)
+    response
+  }
 
   /**
     * Requests get X-Forwarded-For and other request headers added before passing to the service
@@ -66,6 +84,6 @@ case class SecureHeaderFilter(requestHeaders: HeaderMap = SecureHeaders.request,
   def apply(req: Request, service: Service[Request, Response]): Future[Response] =
     for {
       resp <- service(injectRequestHeaders(req))
-      _ <- (resp.headerMap ++= responseHeaders).toFuture
+      _ <- (injectResponseHeaders(resp)).toFuture
     } yield resp
 }
--- a/security/src/main/scala/com/lookout/borderpatrol/security/SecureHeaders.scala
+++ b/security/src/main/scala/com/lookout/borderpatrol/security/SecureHeaders.scala
@@ -14,7 +14,6 @@ import com.twitter.util.Future
   */
 object SecureHeaders {
 
-  val allowOrigin = "Access-Control-Allow-Origin"
   // response headers https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Response_fields
   val StrictTransportSecurity = ("Strict-Transport-Security", "max-age=31557600")
   val XFrameOptions = ("X-Frame-Options", "DENY")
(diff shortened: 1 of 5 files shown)


security/src/main/scala/com/lookout/borderpatrol/security/SecureHeaders.scala
```diff
@@ -13,16 +14,24 @@ import com.twitter.util.Future
   */
 object SecureHeaders {
 
+  val allowOrigin = "Access-Control-Allow-Origin"
```
maheshkelkar: nit - shouldnt this be after comment?

security/src/main/scala/com/lookout/borderpatrol/security/SecureHeaders.scala
```diff
@@ -49,23 +58,32 @@ object SecureHeaders {
   * @param responseHeaders
   */
 case class SecureHeaderFilter(requestHeaders: HeaderMap = SecureHeaders.request,
-                              responseHeaders: HeaderMap = SecureHeaders.response)
+                              responseHeaders: HeaderMap = SecureHeaders.response,
+                              allowedDomains: Set[InternetDomainName])
     extends SimpleFilter[Request, Response] {
+  val validHostsString = allowedDomains.mkString(",")
```
maheshkelkar: we don't need this val

security/src/main/scala/com/lookout/borderpatrol/security/SecureHeaders.scala
```diff
@@ -49,23 +58,32 @@ object SecureHeaders {
...
+  val controlOrigin = SecureHeaders.allowOrigin -> validHostsString
 
   def injectRequestHeaders(req: Request): Request =
     tap(req) { re =>
       re.headerMap ++= requestHeaders
       re.xForwardedFor_=(re.xForwardedFor.fold(localIp)(s => s"$s, $localIp"))
     }
 
+  private[this] def injectResponseHeaders(response: Response): Response = {
+    response.headerMap ++= responseHeaders
```
maheshkelkar: please use `tap`

security/src/main/scala/com/lookout/borderpatrol/security/SecureHeaders.scala
```diff
@@ -49,23 +58,32 @@ object SecureHeaders {
...
+    response.headerMap.add(controlOrigin._1, controlOrigin._2)
```
maheshkelkar: why 2 steps? why not `respose.headerMap ++= (responseHeaders + controlOrigin)`? Or something similar

security/src/main/scala/com/lookout/borderpatrol/security/SecureHeaders.scala
```diff
@@ -49,23 +58,32 @@ object SecureHeaders {
...
+
   /**
     * Requests get X-Forwarded-For and other request headers added before passing to the service
     * Responses get response headers added before returning to the client
     */
   def apply(req: Request, service: Service[Request, Response]): Future[Response] =
     for {
       resp <- service(injectRequestHeaders(req))
-      _ <- (resp.headerMap ++= responseHeaders).toFuture
+      _ <- (injectResponseHeaders(resp)).toFuture
```
maheshkelkar: I am not sure if this will work, unless ofcourse we manage to modify existing `resp`

security/src/test/scala/com/lookout/borderpatrol/test/security/SecureHeadersSpec.scala
```diff
@@ -1,22 +1,28 @@
...
+  val validDomains = Set("www.yahoo.com", "www.google.com", "localhost")
+  implicit val allowedDomains: Set[InternetDomainName] = validDomains map (InternetDomainName.from(_))
+  val filter = SecureHeaderFilter(requestDefaults, defaults, allowedDomains)
   val service = filter andThen testService(r => true)
 
   it should "inject all of the headers" in {
     val request = Request("/")
-    service(request).results.headerMap.sameElements(defaults) should be(true)
+    val map: HeaderMap = service(request).results.headerMap
+    val catchFalse = defaults.map( m => map.contains(m._1)).collect { case false => 1}
```
maheshkelkar: > defaults.map( m => map.contains(m._1)).collect { case false => 1}
> catchFalse.size should be (0)
can you please simply this

security/src/main/scala/com/lookout/borderpatrol/security/SecureHeaders.scala
```diff
@@ -13,16 +14,24 @@ import com.twitter.util.Future
...
   val XDownloadOptions = ("X-Download-Options", "noopen")
   val XPermittedCrossDomainPolicies = ("X-Permitted-Cross-Domain-Policies", "none")
+  val controlEnabled = "Access-Control-Allow-Credentials" -> "true"
+  val exposeHeaders = "Access-Control-Expose-Headers" -> "ETag,Set-Cookie"
+  val allowMethods = "Access-Control-Allow-Methods" -> "GET, POST, PATCH, PUT, DELETE, OPTIONS, HEAD"
+  val allowHeaders =
+    "Access-Control-Allow-Headers" -> "*,x-csrf-token,x-requested-with,Content-Type,If-Modified-Since,If-None-Match"
+  val maxAge = "Access-Control-Max-Age" -> "86400"
 
   val response = HeaderMap(StrictTransportSecurity, XFrameOptions, XXSSProtection,
```
maheshkelkar: if you convert this to a `def response(allowedDomains: Set...)`

and inside `case class` add a `lazy val responseMap = response(allowedDomains)`. it will simply this a lot

security/src/test/scala/com/lookout/borderpatrol/test/security/SecureHeadersSpec.scala
```diff
@@ -38,6 +44,14 @@ class SecureHeadersSpec extends BorderPatrolSuite {
     val response = Response(Status.Ok)
     response.headerMap.add("X-Download-Options", "arglebargle")
     val s = Service.mk[Request, Response](r => Future.value(response))
-    filter(request, s).results.headerMap.sameElements(defaults) should be(true)
+    filter(request, s).results.headerMap("X-Download-Options") should be(SecureHeaders.XDownloadOptions._2)
```
maheshkelkar: I think we are loosing the number of checks here. See above, if you simplify `response`, then you can remove all of this

security/src/main/scala/com/lookout/borderpatrol/security/SecureHeaders.scala
rikesh-chouhan: That is exactly what is happening. Update to existing.

maheshkelkar:
@rikesh-chouhan can you please add a couple of lines in description?

security/src/test/scala/com/lookout/borderpatrol/test/security/SecureHeadersSpec.scala
rikesh-chouhan: I would rather remove this test. The original test was just checking whether or not the response headers in the received response are the same as the `default` headers.
This test is at least confirming that all the expected headers are in included.

security/src/main/scala/com/lookout/borderpatrol/security/SecureHeaders.scala
rikesh-chouhan: Done

security/src/main/scala/com/lookout/borderpatrol/security/SecureHeaders.scala
rikesh-chouhan: but why?
It is following the same existing pattern. Where do you see the simplification?

security/src/test/scala/com/lookout/borderpatrol/test/security/SecureHeadersSpec.scala
rikesh-chouhan: Again, the check is specific to whether or not the response gets updated - true, and whether or not the specific Option that we expect to overwrite is actually being overwritten or not.
I think it improves on current.

security/src/test/scala/com/lookout/borderpatrol/test/security/SecureHeadersSpec.scala
maheshkelkar: >The original test was just checking whether or not the response headers in the received response are the same as the default headers.
Why do you think this is wrong? I think we should ensure that each response contains these set of headers. With your changes these flags are no more default, because `allowedDomains` varies on each instantiation (as defined in config)

security/src/main/scala/com/lookout/borderpatrol/security/SecureHeaders.scala
maheshkelkar: I am not suggesting we change any other code. I am only referring to `val response`. My point was 2 fold. 
1. It will simplify tests
2. You don't have to insert allowedDomains everytime on each request

security/src/test/scala/com/lookout/borderpatrol/test/security/SecureHeadersSpec.scala
maheshkelkar: It looks to me that earlier check is verifying that each element in defaults is included in the response. You modified to check if `X-Download-Options` exists or not. So it looks like you reduced it from many-to-one

security/src/main/scala/com/lookout/borderpatrol/security/SecureHeaders.scala
```diff
@@ -49,23 +59,30 @@ object SecureHeaders {
...
 
   def injectRequestHeaders(req: Request): Request =
     tap(req) { re =>
       re.headerMap ++= requestHeaders
       re.xForwardedFor_=(re.xForwardedFor.fold(localIp)(s => s"$s, $localIp"))
     }
 
+  private[this] def injectResponseHeaders(response: Response): Response =
+    tap(response) { resp =>
+      resp.headerMap ++= (responseHeaders + controlOrigin)
```
maheshkelkar: BTW we don't need to do this operation: `responseHeaders + controlOrigin` while handling each request. You need to do that only once i.e. when `SecureHedaerFilter` is instantiated.

security/src/main/scala/com/lookout/borderpatrol/security/SecureHeaders.scala
rikesh-chouhan: Yes, I agree, that would simplify it, but, I don't want the data living in 2 places. I'll keep the definitions in the current location, but, not add any of the CORS params to the header map. But, then again this would impact the tests.

security/src/test/scala/com/lookout/borderpatrol/test/security/SecureHeadersSpec.scala
rikesh-chouhan: No. This specific test is for  "override existing headers", so the check is to make sure that X-Download-Options has the value from default headers, you see the difference?

There is another test which checks for "all" headers. Which will get updated when I move the params.

security/src/main/scala/com/lookout/borderpatrol/security/SecureHeaders.scala
rikesh-chouhan: This will change the definition of SecureHeaderFilter case class.

security/src/test/scala/com/lookout/borderpatrol/test/security/SecureHeadersSpec.scala
maheshkelkar: ok. got it. Could you please get rid of `._1` or `._2`. I am told by Scala experts that its a strict no-no.

PR #185, opened 2016-08-03
Log and return 403, in case AAD returns anything other than oAuth2 code

### LINKED ISSUE #184: If AAD fails to send oAuth2 code in the redirect after authentication, then return forbidden response
BP requests oAuth2 code while redirecting the user to AAD (OAuth2 server). But, in the redirect after authentication, AAD may send `error` instead of a oAuth2 code. 

Currently, BP redirects the user back to login page. But, that may put it in an infinite loop. Instead BP should log the error and return FORBIDDEN back to the user

maheshkelkar:
- also allow "space" in the query param

Fixes #184

### DIFF
--- a/auth/src/main/scala/com/lookout/borderpatrol/auth/tokenmaster/LoginManagers.scala
+++ b/auth/src/main/scala/com/lookout/borderpatrol/auth/tokenmaster/LoginManagers.scala
@@ -1,9 +1,7 @@
 package com.lookout.borderpatrol.auth.tokenmaster
 
 import java.net.URL
-
-import com.lookout.borderpatrol.auth.{BorderRequest, BpInvalidRequest, SessionIdRequest}
-import com.lookout.borderpatrol.sessionx.Session
+import com.lookout.borderpatrol.auth.{BorderRequest, BpForbiddenRequest, BpInvalidRequest}
 import com.lookout.borderpatrol.{Endpoint, LoginManager}
 import com.lookout.borderpatrol.util.Helpers
 import com.lookout.borderpatrol.util.Combinators.tap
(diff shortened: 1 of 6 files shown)


auth/src/main/scala/com/lookout/borderpatrol/auth/tokenmaster/LoginManagers.scala
```diff
@@ -59,12 +57,18 @@ object LoginManagers {
     def codeToToken(req: BorderRequest): Future[Response] = {
       val hostStr = req.req.host.getOrElse(throw BpInvalidRequest(s"Host not found in HTTP $req"))
```
rikesh-chouhan: Isn't this an HTTP or HTTPS request, might confuse user to think somehow there is some unsecured connection. You can just remove HTTP from that message.

maheshkelkar: ok, done

auth/src/main/scala/com/lookout/borderpatrol/auth/tokenmaster/LoginManagers.scala
```diff
@@ -59,12 +57,18 @@ object LoginManagers {
...
       val scheme = req.req.headerMap.getOrElse("X-Forwarded-Proto", "http")
```
rikesh-chouhan: is this correct? Should it be https?

maheshkelkar: No. "http" is default. Otherwise we take a value from "X-Forwarded-Proto"

rikesh-chouhan:
+2

Issue #186, opened 2016-08-03
Generate access logs


rikesh-chouhan:
Currently borderpatrol does not write and access logs. It should provide a log for the same.

PR #187, opened 2016-08-03
Generate Access Logs

malikvivek:
Issue: #186 

Borderpatrol generates access logs.

### DIFF
--- a/core/src/main/scala/com/lookout/borderpatrol/sessionx/SignedId.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/sessionx/SignedId.scala
@@ -26,6 +26,7 @@ import scala.util.{Success, Failure, Try}
  * @param signature the bytes of the signature(expires, entropy, secret.id)
  */
 case class SignedId(expires: Time, entropy: Entropy, secret: Secret, tag: Tag, signature: Signature) {
+  /* Displaying only last 8 bytes of SessionId in the logs*/
   def toLogIdString: String = Base64StringEncoder.encode(signature.toArray).takeRight(8)
 }
 
--- a/core/src/main/scala/com/lookout/borderpatrol/sessionx/SignedId.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/sessionx/SignedId.scala
@@ -26,7 +26,7 @@ import scala.util.{Success, Failure, Try}
  * @param signature the bytes of the signature(expires, entropy, secret.id)
  */
 case class SignedId(expires: Time, entropy: Entropy, secret: Secret, tag: Tag, signature: Signature) {
-  def toLogIdString: String = Base64StringEncoder.encode(signature.toArray)
+  def toLogIdString: String = Base64StringEncoder.encode(signature.toArray).takeRight(8)
 }
 
 object SignedId {
--- a/example/src/main/scala/com/lookout/borderpatrol/example/service.scala
+++ b/example/src/main/scala/com/lookout/borderpatrol/example/service.scala
@@ -100,7 +100,7 @@ object service {
       case _ =>
         serviceChainFront andThen
           /* Generate the Access Log */
-          AccessLogFilter("bp-accesslogs","bp-access.log", (1*1024*1000)) andThen
+          AccessLogFilter("bp_accesslogs","/var/log/bp_access.log", (1*1024*1024)) andThen
           /* Validate that its our service */
           CustomerIdFilter(serviceMatcher) andThen
           /* Get or allocate Session/SignedId */
--- a/example/src/main/scala/com/lookout/borderpatrol/example/service.scala
+++ b/example/src/main/scala/com/lookout/borderpatrol/example/service.scala
@@ -24,6 +24,7 @@
 package com.lookout.borderpatrol.example
 
 import java.net.URL
+
 import com.lookout.borderpatrol.auth.tokenmaster.LoginManagers.{BasicLoginManager, OAuth2LoginManager}
 import com.lookout.borderpatrol.security.HostHeaderFilter
 import com.lookout.borderpatrol.{HealthCheckRegistry, ServiceMatcher}
@@ -31,7 +32,7 @@ import com.lookout.borderpatrol.auth._
 import com.lookout.borderpatrol.auth.tokenmaster.Tokenmaster._
 import com.lookout.borderpatrol.auth.tokenmaster._
 import com.lookout.borderpatrol.auth.tokenmaster.Tokens._
-import com.lookout.borderpatrol.server.HealthCheckService
+import com.lookout.borderpatrol.server.{AccessLogFilter, HealthCheckService}
 import com.lookout.borderpatrol.sessionx._
 import com.lookout.borderpatrol.util.Combinators._
 import com.twitter.finagle.http.path.Path
@@ -42,6 +43,7 @@ import com.twitter.finagle.http.service.RoutingService
 import com.twitter.finagle.{Filter, Service}
 import com.twitter.util.Future
 import io.circe._
+
 import scala.util.{Failure, Success, Try}
 
 
@@ -97,6 +99,8 @@ object service {
 
       case _ =>
         serviceChainFront andThen
+          /*Generate the Access Log*/
+          AccessLogFilter("bp-accesslogs","bp-access.log", (1*1024*1000)) andThen
           /* Validate that its our service */
           CustomerIdFilter(serviceMatcher) andThen
           /* Get or allocate Session/SignedId */
--- a/example/src/main/scala/com/lookout/borderpatrol/example/service.scala
+++ b/example/src/main/scala/com/lookout/borderpatrol/example/service.scala
@@ -99,7 +99,7 @@ object service {
 
       case _ =>
         serviceChainFront andThen
-          /*Generate the Access Log*/
+          /* Generate the Access Log */
           AccessLogFilter("bp-accesslogs","bp-access.log", (1*1024*1000)) andThen
           /* Validate that its our service */
           CustomerIdFilter(serviceMatcher) andThen
--- a/server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
+++ b/server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
@@ -1,5 +1,6 @@
 package com.lookout.borderpatrol.server
 
+import com.lookout.borderpatrol.util.Combinators.tap
 import com.twitter.util._
 import com.twitter.finagle.{Service, SimpleFilter}
 import com.twitter.finagle.http.{Request, Response}
@@ -17,14 +18,6 @@ import com.twitter.util.Future
 case class AccessLogFilter(name: String, path: String, fileSize: Long)
   extends SimpleFilter[Request, Response] {
 
-  val logger: Logger = {
-    val logger = Logger.get(name)
-    logger.clearHandlers()
-    logger.setLevel(Logger.INFO)
-    logger.setUseParentHandlers(false)
-    logger
-  }
-
   lazy val accessLogHandler = FileHandler(
     filename = path,
     rollPolicy = Policy.MaxSize(new StorageUnit(fileSize)),
@@ -33,24 +26,14 @@ case class AccessLogFilter(name: String, path: String, fileSize: Long)
     formatter = BareFormatter
   ).apply()
 
-  logger.addHandler(new QueueingHandler(accessLogHandler))
-
-  /**
-    * Following code snippet provides a mechanism to replace the logging to a file by logging to
-    * either string or console using the local logger implemented above. Replace the require handler
-    * and pass it while adding the handler to a new QueueingHandler.
-    */
-
-  /**
-    * lazy val local = new Local[String]
-    * lazy val formatter = new Formatter {
-    *   override def format(record: javalog.LogRecord) =
-    *    local().getOrElse("MISSING!:") + formatText(record) + lineTerminator
-    * }
-    * lazy val stringHandler = new StringHandler(formatter, Some(Logger.INFO))
-    * lazy val consoleHandler = new ConsoleHandler(BareFormatter, Some(Logger.INFO))
-    * local() = "I"
-  */
+  val logger: Logger = {
+    tap(Logger.get(name)) { l =>
+      l.clearHandlers()
+      l.setLevel(Logger.INFO)
+      l.setUseParentHandlers(false)
+      l.addHandler(new QueueingHandler(accessLogHandler))
+    }
+  }
 
   def apply(req: Request, service: Service[Request, Response]): Future[Response] = {
     val startTime = Time.now
--- a/server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
+++ b/server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
@@ -1,22 +1,13 @@
 package com.lookout.borderpatrol.server
 
-import java.util.{logging => javalog}
-
-import com.lookout.borderpatrol.auth.{BorderRequest, BpIdentityProviderError,
-BpInvalidRequest, CustomerIdRequest, SessionIdRequest}
-import com.lookout.borderpatrol.sessionx.{BpSignedIdError, SecretStoreApi, Session, SessionStore, SignedId}
-import com.lookout.borderpatrol.{BpNotFoundRequest, CustomerIdentifier, Endpoint,
-LoginManager, ServiceIdentifier, ServiceMatcher}
 import com.twitter.util._
-import com.twitter.finagle.{Filter, Service, SimpleFilter}
-import com.twitter.finagle.http.{Cookie, Method, Request, Response}
+import com.twitter.finagle.{Service, SimpleFilter}
+import com.twitter.finagle.http.{Request, Response}
 import com.twitter.logging.{Logger, _}
 import com.twitter.util.Future
 
-import scala.util.{Failure, Try}
-
 /**
-  * Filter to generate access logs
+  * Filter to generate access logs. The logs generated
   *
   * @param name
   * @param path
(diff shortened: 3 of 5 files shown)


example/src/main/scala/com/lookout/borderpatrol/example/service.scala
```diff
@@ -97,6 +99,8 @@ object service {
 
       case _ =>
         serviceChainFront andThen
+          /*Generate the Access Log*/
```
maheshkelkar: nit - space after comment `/*`

example/src/main/scala/com/lookout/borderpatrol/example/service.scala
```diff
@@ -97,6 +99,8 @@ object service {
...
+          AccessLogFilter("bp-accesslogs","bp-access.log", (1*1024*1000)) andThen
```
maheshkelkar: shouldnt path be `/var/log/bp.access.log`? I think we should use `.` or `_` instead of `-` in the filename

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
```diff
@@ -0,0 +1,70 @@
+package com.lookout.borderpatrol.server
+
+import java.util.{logging => javalog}
+
+import com.lookout.borderpatrol.auth.{BorderRequest, BpIdentityProviderError,
+BpInvalidRequest, CustomerIdRequest, SessionIdRequest}
```
maheshkelkar: I don't think we need most of the imports here

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
```diff
@@ -0,0 +1,70 @@
...
+import com.lookout.borderpatrol.sessionx.{BpSignedIdError, SecretStoreApi, Session, SessionStore, SignedId}
+import com.lookout.borderpatrol.{BpNotFoundRequest, CustomerIdentifier, Endpoint,
```
maheshkelkar: I don't think we need these imports here. Look for `unused imports` in compiler log or take them out if `greyed out` in Intellij

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  val logger: Logger = {
+    val logger = Logger.get(name)
+    logger.clearHandlers()
+    logger.setLevel(Logger.INFO)
+    logger.setUseParentHandlers(false)
+    logger
+  }
+
+  lazy val local = new Local[String]
```
maheshkelkar: why do we need this `val local`? Why can we use `new Local[String]` below?

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
```diff
@@ -0,0 +1,70 @@
...
+  lazy val formatter = new Formatter {
+    override def format(record: javalog.LogRecord) =
+      local().getOrElse("MISSING!:") + formatText(record) + lineTerminator
```
maheshkelkar: What is the purpose of this local()? does it return Option[String]? If you choose to keep the `val`, then please add data type there.

Please add some text (one-liner) to describe the purpose of this `format` method

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  lazy val accessLogHandler = FileHandler(
+    filename = path,
+    rollPolicy = Policy.MaxSize(new StorageUnit(fileSize)),
+    level = Some(Level.INFO),
+    append = true,
+    formatter = BareFormatter
+  ).apply()
+  lazy val stringHandler = new StringHandler(formatter, Some(Logger.INFO))
```
maheshkelkar: this seems unused. In that case we don't need the `formatter` method

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
```diff
@@ -0,0 +1,70 @@
...
+  lazy val consoleHandler = new ConsoleHandler(BareFormatter, Some(Logger.INFO))
+  logger.addHandler(new QueueingHandler(accessLogHandler))
```
maheshkelkar: why not do this up-top while initializing the `val logger`

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  local() = "I"
```
maheshkelkar: local is defined as a val, we are assigning a value here. What is the purpose of this?

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  def apply(req: Request, service: Service[Request, Response]): Future[Response] = {
+
+    val startTime = Time.now
+    for {
+      resp <- service(req)
+      _<- Future(logger.info(s"${req.xForwardedFor.getOrElse("-")}\t${startTime.format("[yyyy/mm/dd:hh:mm:ss.sss]")}\t"+
```
maheshkelkar: I would add a comment to describe the format here.
The add a comment before each print to describe what its prinitng. 
E.g
// Access Log format: AccessLog<version>, ip, request method, request host, etc.

```
logger.info(
  /* ip address */
  req.xForwardedFor.getOrElse....,
  /* time */
  startTime.format...
  /* method */
  req.method
  ::
```

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
```diff
@@ -0,0 +1,70 @@
...
+        s"${req.method}\t${req.host.getOrElse("-")}\t${req.path.getOrDefault("-")}\t" +
+        s"${req.cookies.getValue("border_session").getOrElse("-").takeRight(8)}\t"+
```
maheshkelkar: Instead of 
`req.cookies.getValue("border_session").getOrElse("-").takeRight(8)`
Try
req.cookies.getValue("border_session").fold("-")(c => c.takeRight(8))

server/src/test/scala/com/lookout/borderpatrol/server/AccessLogFilterSpec.scala
```diff
@@ -0,0 +1,46 @@
+/**
+  * Created by vivek.malik on 8/1/16.
+  */
+package com.lookout.borderpatrol.server
+
+import java.util.{Calendar, Date, logging => javalog}
+
+import com.lookout.borderpatrol.auth.{CustomerIdFilter, CustomerIdRequest, SessionIdFilter, SessionIdRequest}
+//import com.lookout.borderpatrol.service.ServiceConfig._
+//import com.lookout.borderpatrol.test.service.serviceTestHelpers._
```
maheshkelkar: please remove dead code

server/src/test/scala/com/lookout/borderpatrol/server/AccessLogFilterSpec.scala
```diff
@@ -0,0 +1,46 @@
+/**
+  * Created by vivek.malik on 8/1/16.
+  */
+package com.lookout.borderpatrol.server
+
+import java.util.{Calendar, Date, logging => javalog}
+
+import com.lookout.borderpatrol.auth.{CustomerIdFilter, CustomerIdRequest, SessionIdFilter, SessionIdRequest}
```
maheshkelkar: I don;t think you need all of these imports. please remove that

server/src/test/scala/com/lookout/borderpatrol/server/AccessLogFilterSpec.scala
```diff
@@ -0,0 +1,46 @@
+/**
+  * Created by vivek.malik on 8/1/16.
+  */
+package com.lookout.borderpatrol.server
```
maheshkelkar: I think package name should be `com.lookout.borderpatrol.test.server`. Compare it with other files

example/src/main/scala/com/lookout/borderpatrol/example/service.scala
malikvivek: /var/log/ causes the tests to fail. We'll remove it and make it configurable in next build. For now to have the feature up, just the file name.

server/src/test/scala/com/lookout/borderpatrol/server/AccessLogFilterSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+class AccessLogFilterSpec extends BorderPatrolSuite {
+  import coreTestHelpers._
+
+  behavior of "AccessLogFilter"
+
+  it should "successfully create a local file and log" in {
+
+    val name = "localAccessLogger"
+    val tempValidFile = File.makeTemp("TempAccessLogFile", ".tmp")
+    val testFileSize: Long = 1*1024*1000
```
maheshkelkar: why not 1*1024*1024?

server/src/test/scala/com/lookout/borderpatrol/server/AccessLogFilterSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+      }
+    }
+
+    // Execute
+    val output = (AccessLogFilter(name, tempValidFile.toCanonical.toString, testFileSize) andThen testService) (Request)
+
+    // Validate
+    Await.result(output).status should be(Status.Ok)
+    Thread.sleep(5)
+    Source.fromFile(tempValidFile.toCanonical.toString).mkString should include("AccessLogV1")
```
maheshkelkar: You may want to match entire line, so that if anybody changes format in future, they can be sure to update the test case

core/src/main/scala/com/lookout/borderpatrol/sessionx/SignedId.scala
```diff
@@ -26,7 +26,7 @@ import scala.util.{Success, Failure, Try}
  * @param signature the bytes of the signature(expires, entropy, secret.id)
  */
 case class SignedId(expires: Time, entropy: Entropy, secret: Secret, tag: Tag, signature: Signature) {
-  def toLogIdString: String = Base64StringEncoder.encode(signature.toArray)
```
maheshkelkar: I had forgotten to add a comment here. Could you please take this opportunity to add a comment saying that instead of exposing entire sessionId in the log file, exposing trailing 8 bytes

example/src/main/scala/com/lookout/borderpatrol/example/service.scala
PeterFCaswell: Yeah, I'd get nervous about absolute path names in the code. It should be something configurable.

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
malikvivek: This local is required as a part of code snippet that uses the local logger for logging on console using consoleHandler or StringHandler.

I've edited the unused code inside a comment and added a comment description for this code. Changes to this might come in handy when we test logs on Sumologic.

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
malikvivek: Same as previous. Local logger testing on console for simultaneous logs.

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
malikvivek: Same as previous. Local logger testing on console for simultaneous logs.

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
malikvivek: Moved it up top and commented out the test snippets.

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
malikvivek: Same as previous. Local logger testing on console for simultaneous logs.

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
malikvivek: Changes noted and made.

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
malikvivek: Removed the unused imports.

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
malikvivek: Changed based upon suggestion.

server/src/test/scala/com/lookout/borderpatrol/server/AccessLogFilterSpec.scala
malikvivek: Discussed and changed to check more fields. All are not possible because they are dynamic and added at run-time.

server/src/test/scala/com/lookout/borderpatrol/server/AccessLogFilterSpec.scala
malikvivek: Because that's how StorageUnit object's defined their sizes as

server/src/test/scala/com/lookout/borderpatrol/server/AccessLogFilterSpec.scala
malikvivek: It is ```com.lookout.borderpatrol.server``` in other files.

core/src/main/scala/com/lookout/borderpatrol/sessionx/SignedId.scala
malikvivek: Comment added

example/src/main/scala/com/lookout/borderpatrol/example/service.scala
```diff
@@ -97,6 +99,8 @@ object service {
 
       case _ =>
         serviceChainFront andThen
+          /* Generate the Access Log */
+          AccessLogFilter("bp-accesslogs","bp-access.log", (1*1024*1000)) andThen
```
maheshkelkar: 1. "/var/log/bp.access.log" - eventyally comming from config
2. why `1*1024*1000`

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
```diff
@@ -0,0 +1,86 @@
...
+
+  lazy val accessLogHandler = FileHandler(
+    filename = path,
+    rollPolicy = Policy.MaxSize(new StorageUnit(fileSize)),
+    level = Some(Level.INFO),
+    append = true,
+    formatter = BareFormatter
+  ).apply()
+
+  logger.addHandler(new QueueingHandler(accessLogHandler))
```
maheshkelkar: why not move this at line 25?

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
```diff
@@ -0,0 +1,86 @@
...
+  /**
+    * lazy val local = new Local[String]
+    * lazy val formatter = new Formatter {
+    *   override def format(record: javalog.LogRecord) =
+    *    local().getOrElse("MISSING!:") + formatText(record) + lineTerminator
+    * }
+    * lazy val stringHandler = new StringHandler(formatter, Some(Logger.INFO))
+    * lazy val consoleHandler = new ConsoleHandler(BareFormatter, Some(Logger.INFO))
+    * local() = "I"
+  */
```
maheshkelkar: I think you can move this to test code if you want. This is an example code, I am not sure if we should keep it here.

server/src/test/scala/com/lookout/borderpatrol/server/AccessLogFilterSpec.scala
```diff
@@ -0,0 +1,40 @@
+package com.lookout.borderpatrol.server
```
maheshkelkar: package?

server/src/test/scala/com/lookout/borderpatrol/server/AccessLogFilterSpec.scala
maheshkelkar: StorageUnit simply takes nbr of bytes, it doesnt care if its MB or KB

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
malikvivek: The code has been moved to local branch for future reference.

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
malikvivek: I'm not sure what you meant here; but if you're suggesting to move line 36 to line 25, it needs accessLogHandler as a parameter.

example/src/main/scala/com/lookout/borderpatrol/example/service.scala
malikvivek: Size moved to 1024*1024.

rikesh-chouhan:
+2

PR #189, opened 2016-08-05
If destination query param is URL, then validate that host belongs to Set of valid hosts

### LINKED ISSUE #188: Check whether or not log out destination is valid
Determine whether or not the a "landing" parameter - for example "destination" - is an allowed host name. 
If the value is a URL then and only then get host value and check if it is an allowed host.
YES - proceed to that url
NO - return NONE to allow the caller to take action accordingly.

If the value is not a URL return as is

rikesh-chouhan:
Push changes to check if host name for a "destination" is allowed.
Caller can proceed on action with value obtained from DestinationValidator

Fixes #188

### DIFF
--- a/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
@@ -397,11 +397,11 @@ case class LogoutService(store: SessionStore, destinationValidator: DestinationV
       log.debug(s"Logging out Session: ${sid.toLogIdString}")
       store.delete(sid)
     })
-    // Redirect to (1) the logged out url, (2) suggested url or (3) default service path, in that order
-    val location: String = (req.customerId.loginManager.loggedOutUrl,
-      BorderAuth.extractDestination(req.req.params)(destinationValidator)) match {
+    // Redirect to (1) suggested url or (2) the logged out url or (3) default service path, in that order
+    val location: String = (BorderAuth.extractDestination(req.req.params)(destinationValidator),
+      req.customerId.loginManager.loggedOutUrl) match {
       case (Some(loc), _) => loc.toString
-      case (None, Some(loc)) => loc
+      case (Some(loc),None) => loc
       case _ => {
         log.info("Could not determine host: "+req.req.params.get("destination")+" using default")
         req.customerId.defaultServiceId.path.toString
@@ -410,7 +410,6 @@ case class LogoutService(store: SessionStore, destinationValidator: DestinationV
     BorderAuth.formatLogoutResponse(req.req, Status.Ok, location,
       s"After logout, redirecting to: '$location'").toFuture
   }
-
 }
 
 /**
--- a/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
@@ -254,8 +254,13 @@ case class SendToIdentityProvider(identityProviderMap: Map[String, Service[Borde
        */
       case (None, _)
         if Path(req.req.path).startsWith(req.customerId.loginManager.loginConfirm) => {
-        val location = BorderAuth.attemptDestinationFetch(req.req.params)(destinationValidator)
-          .getOrElse(req.customerId.defaultServiceId.path.toString)
+        val location = BorderAuth.attemptDestinationFetch(req.req.params)(destinationValidator) match {
+          case Some(s) => s
+          case _ => {
+            log.info("Could not determine host: "+req.req.params.get("destination")+" using default")
+            req.customerId.defaultServiceId.path.toString
+          }
+        }
         for {
           sessionId <- SignedId.untagged
           session <- Session(sessionId, Request(location)).toFuture
--- a/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
@@ -195,8 +195,8 @@ case class SessionIdFilter(matcher: ServiceMatcher, store: SessionStore)(
  * @param secretStore
  */
 case class SendToIdentityProvider(identityProviderMap: Map[String, Service[BorderRequest, Response]],
-                                  store: SessionStore, destinationValidator: DestinationValidator)(
-  implicit secretStore: SecretStoreApi, statsReceiver: StatsReceiver)
+                                  store: SessionStore)(
+  implicit destinationValidator: DestinationValidator, secretStore: SecretStoreApi, statsReceiver: StatsReceiver)
     extends SimpleFilter[SessionIdRequest, Response] {
   private[this] val log = Logger.get(getClass.getPackage.getName)
   private[this] val statLoginRedirects = statsReceiver.counter("req.login.required.redirects")
@@ -387,8 +387,8 @@ case class SendToUnprotectedService(store: SessionStore)
  * - sets the empty cookie in response
  * - redirects to default service path
  */
-case class LogoutService(store: SessionStore, destinationValidator: DestinationValidator)
-                        (implicit secretStore: SecretStoreApi)
+case class LogoutService(store: SessionStore)(implicit destinationValidator: DestinationValidator,
+                        secretStore: SecretStoreApi)
     extends Service[CustomerIdRequest, Response] {
   private[this] val log = Logger.get(getClass.getPackage.getName)
 
@@ -398,7 +398,7 @@ case class LogoutService(store: SessionStore, destinationValidator: DestinationV
       store.delete(sid)
     })
     // Redirect to (1) suggested url or (2) the logged out url or (3) default service path, in that order
-    val location: String = (BorderAuth.extractDestination(req.req.params)(destinationValidator),
+    val location: String = (BorderAuth.extractDestination(req.req.params),
       req.customerId.loginManager.loggedOutUrl) match {
       case (Some(loc), _) => loc.toString
       case (Some(loc),None) => loc
--- a/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
+++ b/core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
@@ -114,6 +114,21 @@ object BorderAuth {
         else r.uri.replaceFirst(serviceId.path.toString, rwPath.toString)
       }
     }
+
+  /**
+    * Centralize attempt to fetch a 'destination' param. In future this could be parametrized.
+    * Try to get a scrubbed value that then needs to be checked against known host names.
+    *
+    * @param paramMap
+    * @param destinationValidator
+    * @return Option if a valid hostname (value) was extracted from paramMap
+    */
+  def attemptDestinationFetch(paramMap: ParamMap)(implicit destinationValidator: DestinationValidator): Option[String] =
+    for {
+      loc <- Helpers.scrubQueryParams(paramMap, "destination")
+      a <- destinationValidator.matchesValidHosts(loc)
+    } yield a
+
 }
 
 /**
@@ -180,7 +195,7 @@ case class SessionIdFilter(matcher: ServiceMatcher, store: SessionStore)(
  * @param secretStore
  */
 case class SendToIdentityProvider(identityProviderMap: Map[String, Service[BorderRequest, Response]],
-                                  store: SessionStore)(
+                                  store: SessionStore, destinationValidator: DestinationValidator)(
   implicit secretStore: SecretStoreApi, statsReceiver: StatsReceiver)
     extends SimpleFilter[SessionIdRequest, Response] {
   private[this] val log = Logger.get(getClass.getPackage.getName)
@@ -239,7 +254,7 @@ case class SendToIdentityProvider(identityProviderMap: Map[String, Service[Borde
        */
       case (None, _)
         if Path(req.req.path).startsWith(req.customerId.loginManager.loginConfirm) => {
-        val location = Helpers.scrubQueryParams(req.req.params, "destination")
+        val location = BorderAuth.attemptDestinationFetch(req.req.params)(destinationValidator)
           .getOrElse(req.customerId.defaultServiceId.path.toString)
         for {
           sessionId <- SignedId.untagged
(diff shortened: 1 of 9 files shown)


PeterFCaswell:
+1
Though I'm kind of confused why we're "scrubbing" some strings.

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -377,15 +393,19 @@ case class LogoutService(store: SessionStore)(implicit secretStore: SecretStoreA
       store.delete(sid)
     })
     // Redirect to (1) the logged out url, (2) suggested url or (3) default service path, in that order
-    val location = (req.customerId.loginManager.loggedOutUrl,
-      Helpers.scrubQueryParams(req.req.params, "destination")) match {
+    val location: String = (req.customerId.loginManager.loggedOutUrl,
+      BorderAuth.attemptDestinationFetch(req.req.params)(destinationValidator)) match {
```
rikesh-chouhan: @kgangwani - The code here is attempting to do a best effort param value retrieval, is that enough? logging for failure etc, is there anything else that needs to occur?

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -239,7 +254,7 @@ case class SendToIdentityProvider(identityProviderMap: Map[String, Service[Borde
        */
       case (None, _)
         if Path(req.req.path).startsWith(req.customerId.loginManager.loginConfirm) => {
-        val location = Helpers.scrubQueryParams(req.req.params, "destination")
+        val location = BorderAuth.attemptDestinationFetch(req.req.params)(destinationValidator)
```
rikesh-chouhan: Best effort. Is this sufficient? I'll add a log statement.

core/src/main/scala/com/lookout/borderpatrol/auth/DestinationValidator.scala
```diff
@@ -0,0 +1,38 @@
...
+    * If yes - return it wrapped as Option
+    * If no -
+    *   Is this a URL
+    *   Yes - Attempt to get host and find if it is a valid host entry
+    *   No - (could be relative path) return the fragment as is
+    *
+    * @param location (could be a URL - if yes attempt to extract host from entry and use that for match)
+    * @return
+    */
+  def matchesValidHosts(location: String): Option[String] =
```
maheshkelkar: can we change the method name to `matchHost` or `checkHost`, because thats what we are going

core/src/main/scala/com/lookout/borderpatrol/auth/DestinationValidator.scala
```diff
@@ -0,0 +1,38 @@
+package com.lookout.borderpatrol.auth
```
maheshkelkar: I wonder if this should be in the `Util` package, just like scrubbing code

core/src/main/scala/com/lookout/borderpatrol/auth/DestinationValidator.scala
```diff
@@ -0,0 +1,38 @@
...
+    * If no -
+    *   Is this a URL
+    *   Yes - Attempt to get host and find if it is a valid host entry
+    *   No - (could be relative path) return the fragment as is
+    *
+    * @param location (could be a URL - if yes attempt to extract host from entry and use that for match)
+    * @return
+    */
+  def matchesValidHosts(location: String): Option[String] =
+    Try(new URL(location)) match {
```
maheshkelkar: nit - space after `Try`

core/src/main/scala/com/lookout/borderpatrol/util/Helpers.scala
```diff
@@ -14,7 +14,11 @@ object Helpers {
       /* These param values could be malformed and may contain special characters. So lets scrub them out and
        * choose the first valid string as param
        */
-      l.split(specialCharRegEx).filterNot(_.isEmpty).headOption.map(_.trim)
+      scrubAString(l)
     }
   }
+
+  def scrubAString(toScrub: String): Option[String] = {
```
maheshkelkar: why?

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -114,6 +114,21 @@ object BorderAuth {
...
+
+  /**
+    * Centralize attempt to fetch a 'destination' param. In future this could be parametrized.
+    * Try to get a scrubbed value that then needs to be checked against known host names.
+    *
+    * @param paramMap
+    * @param destinationValidator
+    * @return Option if a valid hostname (value) was extracted from paramMap
+    */
+  def attemptDestinationFetch(paramMap: ParamMap)(implicit destinationValidator: DestinationValidator): Option[String] =
```
maheshkelkar: I am not sure why do we need this method here? How about incorporating this scrubQueryParams in the DestinationValidator itself?

Also, if you insist on this, we should rename this method to something simple.

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -239,8 +254,13 @@ case class SendToIdentityProvider(identityProviderMap: Map[String, Service[Borde
        */
       case (None, _)
         if Path(req.req.path).startsWith(req.customerId.loginManager.loginConfirm) => {
-        val location = Helpers.scrubQueryParams(req.req.params, "destination")
-          .getOrElse(req.customerId.defaultServiceId.path.toString)
+        val location = BorderAuth.attemptDestinationFetch(req.req.params)(destinationValidator) match {
```
maheshkelkar: we don't need to pass `destinationValidator` if its implicit

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
maheshkelkar: we don't need to pass `destinationValidator` if its implicit

rikesh-chouhan: I was getting compile and test failures because destinationValidator was not within scope. 
I think it should be present in the function.

core/src/main/scala/com/lookout/borderpatrol/auth/DestinationValidator.scala
rikesh-chouhan: sure.

core/src/main/scala/com/lookout/borderpatrol/auth/DestinationValidator.scala
rikesh-chouhan: No, IMO. 
Unfortunately it has state. It is initialized. So not a util right now.
When we have a config object that allows us to fetch the hosts dynamically (whatever they maybe) - Yes, it can be util.

core/src/main/scala/com/lookout/borderpatrol/auth/DestinationValidator.scala
rikesh-chouhan: ok.

core/src/main/scala/com/lookout/borderpatrol/util/Helpers.scala
rikesh-chouhan: If you need to scrub just a string in future (very likely) you have readymade util code to do it.

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
rikesh-chouhan: No, I would not incorporate scrubQueryParams into DestinationValidator.
I don't like to assume or tie the value extraction to a specific implementation.
I'll change the name of the function.

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -377,15 +398,19 @@ case class LogoutService(store: SessionStore)(implicit secretStore: SecretStoreA
       store.delete(sid)
     })
     // Redirect to (1) the logged out url, (2) suggested url or (3) default service path, in that order
-    val location = (req.customerId.loginManager.loggedOutUrl,
-      Helpers.scrubQueryParams(req.req.params, "destination")) match {
+    val location: String = (req.customerId.loginManager.loggedOutUrl,
+      BorderAuth.attemptDestinationFetch(req.req.params)(destinationValidator)) match {
       case (Some(loc), _) => loc.toString
       case (None, Some(loc)) => loc
```
maheshkelkar: Can you take this opportunity to swap this? And give higher priority to `destination` query param and then to loggedOutUrl config? Also, please update the comment accordingly

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -377,15 +398,19 @@ case class LogoutService(store: SessionStore)(implicit secretStore: SecretStoreA
...
-      case _ => req.customerId.defaultServiceId.path.toString
+      case _ => {
+        log.info("Could not determine host: "+req.req.params.get("destination")+" using default")
+        req.customerId.defaultServiceId.path.toString
+      }
     }
     BorderAuth.formatLogoutResponse(req.req, Status.Ok, location,
       s"After logout, redirecting to: '$location'").toFuture
   }
+
```
maheshkelkar: nit - please delete newline

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -114,6 +114,21 @@ object BorderAuth {
...
+    * Centralize attempt to fetch a 'destination' param. In future this could be parametrized.
+    * Try to get a scrubbed value that then needs to be checked against known host names.
+    *
+    * @param paramMap
+    * @param destinationValidator
+    * @return Option if a valid hostname (value) was extracted from paramMap
+    */
+  def attemptDestinationFetch(paramMap: ParamMap)(implicit destinationValidator: DestinationValidator): Option[String] =
+    for {
+      loc <- Helpers.scrubQueryParams(paramMap, "destination")
```
maheshkelkar: Another thought; If we keep scrubQueryParams seperate, then DestinationValidator simply reduces to a single check i.e. `validHosts.contains(location)`

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
rikesh-chouhan: Sure.

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
rikesh-chouhan: Since, @kgangwani agrees with current behavior, lets leave this as is.
validHosts.contains also returns a Boolean.

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
rikesh-chouhan: done

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
rikesh-chouhan: Same issue, compile and test failures, not within scope.

core/src/main/scala/com/lookout/borderpatrol/util/Helpers.scala
maheshkelkar: Could you please add seperate comments on top of each method to describe purpose of these 2 methods?

core/src/main/scala/com/lookout/borderpatrol/auth/BorderAuth.scala
```diff
@@ -376,12 +397,15 @@ case class LogoutService(store: SessionStore)(implicit secretStore: SecretStoreA
       log.debug(s"Logging out Session: ${sid.toLogIdString}")
       store.delete(sid)
     })
-    // Redirect to (1) the logged out url, (2) suggested url or (3) default service path, in that order
-    val location = (req.customerId.loginManager.loggedOutUrl,
-      Helpers.scrubQueryParams(req.req.params, "destination")) match {
+    // Redirect to (1) suggested url or (2) the logged out url or (3) default service path, in that order
+    val location: String = (BorderAuth.extractDestination(req.req.params)(destinationValidator),
```
maheshkelkar: `destinationValidator` is implicit, no need to pass it here. Otherwise it should not be implicit

core/src/main/scala/com/lookout/borderpatrol/util/Helpers.scala
```diff
@@ -8,13 +8,26 @@ object Helpers {
...
-      /* These param values could be malformed and may contain special characters. So lets scrub them out and
-       * choose the first valid string as param
-       */
-      l.split(specialCharRegEx).filterNot(_.isEmpty).headOption.map(_.trim)
+      scrubAString(l)
     }
   }
+
+  /**
+    * Scrub a string passed in and return value
```
maheshkelkar: To describe what scrubbing means: Can you copy line from comment above here

core/src/test/scala/com/lookout/borderpatrol/test/auth/BorderAuthSpec.scala
```diff
@@ -1019,7 +1022,7 @@ class BorderAuthSpec extends BorderPatrolSuite {
     request.addCookie(cooki2)
 
     // Execute
-    val output = (ExceptionFilter() andThen CustomerIdFilter(serviceMatcher) andThen LogoutService(sessionStore))(
+    val output = (ExceptionFilter() andThen CustomerIdFilter(serviceMatcher) andThen LogoutService(sessionStore, destinationValidator))(
```
maheshkelkar: if column width is more than 100, then please put this on newline

maheshkelkar:
@rikesh-chouhan: I have posted a few minor comments. Please push a new commit with those changes and I can merge it. Thanks!

rikesh-chouhan:
Removed implicit param from function call and updated tests accordingly.

Issue #193, opened 2016-08-15
Access Logs


malikvivek:
Create a directory for the access log file, if it does not exist.

PR #194, opened 2016-08-15
Access Logs configuration.

malikvivek:
Create log file directory if does not exist.

Issue: #193

### DIFF
--- a/example/src/main/scala/com/lookout/borderpatrol/example/service.scala
+++ b/example/src/main/scala/com/lookout/borderpatrol/example/service.scala
@@ -81,10 +81,12 @@ object service {
     val notFoundService = Service.mk[SessionIdRequest, Response] { req => Response(Status.NotFound).toFuture }
     implicit val destinationValidator = DestinationValidator(config.allowedDomains)
     val serviceChainFront: Filter[Request, Response, Request, Response] =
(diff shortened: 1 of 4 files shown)


example/src/main/scala/com/lookout/borderpatrol/example/service.scala
```diff
@@ -88,13 +88,15 @@ object service {
...
-        ExceptionFilter() andThen
+        AccessLogFilter(config.accessLogConfig.fileName, config.accessLogConfig.fileSizeInMegaBytes) andThen
+          /* Convert exceptions to responses */
+          ExceptionFilter() andThen
           HealthCheckService(registry, BpBuild.BuildInfo.version)
 
       /** Logout */
       case "/logout" =>
         serviceChainFront andThen
+          AccessLogFilter(config.accessLogConfig.fileName, config.accessLogConfig.fileSizeInMegaBytes) andThen
```
maheshkelkar: this is not needed; AccessLogFilter is already part of `serviceChainFront`

malikvivek: This came in due to an undo. Removing it.

rikesh-chouhan:
+2

Issue #190, opened 2016-08-10
Make Access Logs Configurable


malikvivek:
Add configurable parameters for access logs.

PR #195, opened 2016-08-18
Allow Access logs to be optional

malikvivek:
Allow output to file, stdout, stderr
Add fileCount parameter

Issue: #190

### DIFF
--- a/example/src/main/scala/com/lookout/borderpatrol/example/ServerConfig.scala
+++ b/example/src/main/scala/com/lookout/borderpatrol/example/ServerConfig.scala
@@ -11,7 +11,6 @@ import com.twitter.app.App
 import cats.data.Xor
 import com.twitter.finagle.http.path.Path
 import com.twitter.logging.Logger
-import com.twitter.logging.config.Level
 import io.circe.{Encoder, _}
 import io.circe.jawn._
 import io.circe.generic.auto._
--- a/example/src/main/scala/com/lookout/borderpatrol/example/ServerConfig.scala
+++ b/example/src/main/scala/com/lookout/borderpatrol/example/ServerConfig.scala
@@ -11,6 +11,7 @@ import com.twitter.app.App
 import cats.data.Xor
 import com.twitter.finagle.http.path.Path
 import com.twitter.logging.Logger
+import com.twitter.logging.config.Level
 import io.circe.{Encoder, _}
 import io.circe.jawn._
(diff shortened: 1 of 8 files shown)


rikesh-chouhan:
+1.
I'm wondering why the access Log Level is not configurable? Any reason?

malikvivek:
I tried adding level. It picks it as a String and I couldn't figure out it's conversion to level type. Then post discussion with Mahesh, We decided to keep it at INFO and change when required from the finagle console.

rikesh-chouhan:
+2

example/src/main/scala/com/lookout/borderpatrol/example/ServerConfig.scala
```diff
@@ -11,6 +11,7 @@ import com.twitter.app.App
 import cats.data.Xor
 import com.twitter.finagle.http.path.Path
 import com.twitter.logging.Logger
+import com.twitter.logging.config.Level
```
maheshkelkar: we are not using this anymore

example/src/main/scala/com/lookout/borderpatrol/example/service.scala
```diff
@@ -72,6 +71,13 @@ object service {
     "tokenmaster.oauth2" -> tokenmasterAccessIssuerChain(sessionStore)
   )
 
+
+  def accessLogFilter(config: ServerConfig): Filter[Request, Response, Request, Response] =
+    config.accessLogConfig match {
+    case Some(conf) => AccessLogFilter(conf.output, conf.fileSizeInMegaBytes, Level.INFO, conf.fileCount)
```
maheshkelkar: remove Level.INFO

server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
```diff
@@ -40,10 +39,11 @@ case class AccessLogFilter(output: String, fileSizeInMegaBytes: Long, accessLogL
   val logger: Logger = {
     tap(Logger.get(loggerName)) { l =>
       l.clearHandlers()
-      // No need to set log level here so as to prevent the logger from forcing a log level
       // This allows the AccessLog logger to be used as a separate logger, not under root logger.
```
maheshkelkar: Actually it should read:
This allows the AccessLog logger to use different handlers

PR #197, opened 2016-08-25
Bug Fix - Log initial request by caching it and not modified request

malikvivek:
Issue: #193

### DIFF
--- a/server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
+++ b/server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
@@ -49,32 +49,38 @@ case class AccessLogFilter(output: String, fileSizeInMegaBytes: Long, fileCount:
   }
 
   def apply(req: Request, service: Service[Request, Response]): Future[Response] = {
+    val requestIPAddress = req.xForwardedFor.getOrElse("-")
     val startTime = Time.now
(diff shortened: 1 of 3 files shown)


server/src/main/scala/com/lookout/borderpatrol/server/AccessLogFilter.scala
```diff
@@ -49,26 +49,32 @@ case class AccessLogFilter(output: String, fileSizeInMegaBytes: Long, fileCount:
...
+    val requestHost = req.host.getOrElse("-")
+    val requestPath = req.path.getOrDefault("-")
+    val requestSessionId = req.cookies.getValue("border_session").fold("-")(c => c.takeRight(8))
+    val requestContentLength = req.contentLength.getOrElse("-")
     for {
       resp <- service(req)
       _<- Future(logger.log(accessLogLevel,
         /* IP Address */
-        s"${req.xForwardedFor.getOrElse("-")}\t"+
+        s"${requestIPAddress}\t"+
```
PeterFCaswell: Some of these lines have ' +' and some just have '+' at their ends. It would be nice to have consistency.

malikvivek: Noted! Will change

rikesh-chouhan:
+1, if the code is updated for consistency will make it +2

PR #200, opened 2016-09-01
Consolidated and updated AAD token access methods

maheshkelkar:
Also,
- using null tracer (disable zipkin)
- Use default port if no port specified in the URL

### DIFF
--- a/auth/src/main/scala/com/lookout/borderpatrol/auth/tokenmaster/OAuth2.scala
+++ b/auth/src/main/scala/com/lookout/borderpatrol/auth/tokenmaster/OAuth2.scala
@@ -144,7 +144,7 @@ object OAuth2 {
      * @return
      */
     def codeToClaimsSet(req: BorderRequest, loginManager: OAuth2LoginManagerMixin):
-      Future[(JWTClaimsSet, JWTClaimsSet)] = {
+      Future[(String, JWTClaimsSet, JWTClaimsSet)] = {
       for {
         aadToken <- loginManager.codeToToken(req).flatMap(res => res.status match {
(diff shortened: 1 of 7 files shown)


PeterFCaswell:
Looks fine to me. @rikesh-chouhan can you also look through it?

auth/src/main/scala/com/lookout/borderpatrol/auth/tokenmaster/OAuth2.scala
```diff
@@ -144,7 +144,7 @@ object OAuth2 {
      * @return
      */
     def codeToClaimsSet(req: BorderRequest, loginManager: OAuth2LoginManagerMixin):
-      Future[(JWTClaimsSet, JWTClaimsSet)] = {
+      Future[(String, JWTClaimsSet, JWTClaimsSet)] = {
```
rikesh-chouhan: Can we make this a class? The usage of _2 and _1 to access values makes it little ugly.

rikesh-chouhan:
Looks good to me except for 1 portion that I commented on

auth/src/main/scala/com/lookout/borderpatrol/auth/tokenmaster/OAuth2.scala
maheshkelkar: We can, but then we will be ignoring Scala matching capabilities. I mean
`val (a, b) = codeClaimsSet(...)` IMO is better than

As you can see, we are not really `._1` in the code. Thats mainly in the test code, and I can get rid of that one.

PR #201, opened 2016-09-09
Add OptionsFilter to handle HTTP OPTIONS request

maheshkelkar:
Also,
- remove methods from RoutingService
- Disable failfast if upstream is ELB
- Synchronize the Endpoint cache to ensure only 1 instance of Client
  created per endpoint


rikesh-chouhan:
+2

PR #202, opened 2016-09-15
Remove CORS from SecureHeaderFilter and use Cors.HttpFilter instead

maheshkelkar:
- Cors.HttpFilter is available in finagle framework


rikesh-chouhan:
+2

PR #205, opened 2016-09-27
Allow one or more unprotected paths on the main anchor path

maheshkelkar:
- Currently service can be either defined as protected or unprotected. But
  thats not sufficient, as each service may have one or more such paths (pockets)
- By default the anchor path would be protected
- A set of unprotected paths, if specified, will create the necessary holes


rikesh-chouhan:
+2

